PR #32, opened 2023-10-05
Update SDK to use obfuscated RAC

sameerank:
---
labels: mergeable
---
Fixes: [FF-981](https://linear.app/eppo/issue/FF-981/update-js-client-sdk-to-use-obfuscated-rac)

## Motivation and Context

## Description

Bumps the common library dependency and updates tests to confirm that the SDK is working with the obfuscated RAC file

## How has this been tested?

`yarn test`

[//]: # (OPTIONAL)


src/index.spec.ts
```diff
@@ -21,9 +22,10 @@ describe('EppoJSClient E2E test', () => {
   let globalClient: IEppoClient;
 
   const flagKey = 'mock-experiment';
+  const hashedFlagKey = md5(flagKey);
```
sameerank, line 25: A separate variable because we still pass in the un-hashed flag key to get assignments. We use the hashed version for mocking the config store

src/index.ts
```diff
@@ -44,12 +45,90 @@ const localStorage = new EppoLocalStorage();
...
   public static instance: EppoJSClient = new EppoJSClient(localStorage);
+
+  public getAssignment(
+    subjectKey: string,
+    flagKey: string,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    subjectAttributes?: Record<string, any>,
+    assignmentHooks?: IAssignmentHooks,
+  ): string | null {
+    return super.getAssignment(subjectKey, flagKey, subjectAttributes, assignmentHooks, true);
```
sameerank, line 56: We're overriding the client's instance methods because the last argument (`obfuscated: boolean`) isn't meant to be available to the end user. It's determined by the SDK name and version

Makefile
```diff
@@ -34,7 +34,7 @@ test-data:
 	rm -rf $(testDataDir)
 	mkdir -p $(tempDir)
 	git clone -b ${branchName} --depth 1 --single-branch ${githubRepoLink} ${gitDataDir}
-	cp ${gitDataDir}rac-experiments-v3.json ${testDataDir}
+	cp ${gitDataDir}rac-experiments-v3-obfuscated.json ${testDataDir}
```
sameerank, line 37: Replaced to use the obfuscated mock RAC file

test/testHelpers.ts
```diff
@@ -5,7 +5,7 @@ import { IVariation } from '@eppo/js-client-sdk-common/dist/dto/variation-dto';
 
 export const TEST_DATA_DIR = './test/data/';
 export const ASSIGNMENT_TEST_DATA_DIR = TEST_DATA_DIR + 'assignment-v2/';
-export const MOCK_RAC_RESPONSE_FILE = 'rac-experiments-v3.json';
+export const MOCK_RAC_RESPONSE_FILE = 'rac-experiments-v3-obfuscated.json';
```
sameerank, line 8: Matches the filename in the Makefile update

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@eppo/js-client-sdk",
-  "version": "1.2.3",
+  "version": "1.3.0",
```
sameerank, line 3: Minor increment seems right to me since this is meant to work with a differently formatted RAC

ploomiz (approved):

Makefile
ploomiz: I wonder whether the RAC should use a protobuf encoding to reduce size

leoromanovsky: Good idea to consider this - I had forgotten this was possible as last time I used protobufs the encoding was still in json (by accident!); using the native wire format would be a nice way to solve for this.

leoromanovsky (approved):

PR #33, opened 2023-10-18
upgrade to commons 1.7.0; enable non expiring assignment cache by default

leoromanovsky:
Fixes: #__issue__

## Motivation and Context

## Description

By default the client is initialized with assignment cache as the "non-expiring" type

## How has this been tested?

[//]: # (OPTIONAL)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,6 +8,7 @@ import {
   HttpClient,
   IAssignmentHooks,
 } from '@eppo/js-client-sdk-common';
+import { AssignmentCache, Cacheable } from '@eppo/js-client-sdk-common/dist/assignment-cache';
 import axios from 'axios';
 
 import { EppoLocalStorage } from './local-storage';
@@ -33,6 +34,11 @@ export interface IClientConfig {
    * Pass a logging implementation to send variation assignments to your data warehouse.
    */
   assignmentLogger: IAssignmentLogger;
+
+  /**
+   * Pass an assignment cache implementation to cache assignments.
+   */
+  assignmentCache?: AssignmentCache<Cacheable>;
 }
 
 export { IAssignmentLogger, IAssignmentEvent, IEppoClient } from '@eppo/js-client-sdk-common';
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -131,7 +130,7 @@ export class EppoJSClient extends EppoClient {
  * @param config - client configuration
  * @public
  */
-export async function init(config: IClientConfig): Promise<IEppoClient> {
+export async function init(config: IClientConfig): Promise<EppoJSClient> {
```
leoromanovsky, line 133: Might have made a mess of things:

* `IEppoClient` (commons) has the public method definitions - `get*Assignment`
* `EppoClient` - (commons) `setLogger`, `disableAssignmentCache`, etc
* `EppoJSClient` - (this repo) - local configuration

To me it makes some sense to return the most locally relevant client `EppoJSClient`, here.

sameerank: The `EppoJSClient` changes make sense to me too. My main concern was that it would mean that clients would need to update types in their code after updating their SDK version, which is undesirable as far as making updates as low friction as possible

Are there any disadvantages to putting `disableAssignmentCache` in the common library?

aarsilv: Agree with not making a breaking change here.

I do like the the idea of letting them deviate from our defaults should they want to with config options though (even though, like Eric, I don't immediately see a reason why they would other than extreme optimizations).

src/index.ts
```diff
@@ -143,6 +142,11 @@ export async function init(config: IClientConfig): Promise<IEppoClient> {
     sdkVersion,
   });
   EppoJSClient.instance.setLogger(config.assignmentLogger);
+
+  // default behavior is to use a non-expiring cache.
+  // this can be overridden after initialization.
+  EppoJSClient.instance.useNonExpiringAssignmentCache();
```
aarsilv, line 148: :+1:

sameerank (approved):

PR #43, opened 2023-12-18
Adds LocalStorageAssignmentCache with persistent implementation between sessions to lower assignment logging; enabled by default (FF-839)

leoromanovsky:
---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context

Downstream related change: https://github.com/Eppo-exp/js-client-sdk-common/pull/33 enables passing an implementation.

The goal is for users of Eppo's SDK in the browser to remove duplicate assignments within and now between sessions.

We are creating a cache based on `LocalStorage` which is a unique API to browsers.

## Description

* Use only a single `LocalStorage` key for the entire assignment cache - it is always for a single subject.
* Instantiate the js client with this cache.

## How has this been tested?

* New integration test.

[//]: # (OPTIONAL)

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -283,7 +283,24 @@ describe('EppoJSClient E2E test', () => {
           allocationKey: 'allocation-1',
           variationValue: EppoValue.String('control'),
         }),
-      ).toEqual(true);
+      ).toEqual(true); // this key has been logged
+
+      // change variation
+      cache.setLastLoggedAssignment({
+        subjectKey: 'subject-1',
+        flagKey: 'flag-1',
+        allocationKey: 'allocation-1',
+        variationValue: EppoValue.String('variant'),
+      });
+
+      expect(
+        cache.hasLoggedAssignment({
+          subjectKey: 'subject-1',
+          flagKey: 'flag-1',
+          allocationKey: 'allocation-1',
+          variationValue: EppoValue.String('control'),
+        }),
+      ).toEqual(false); // this key has not been logged
     });
   });
 
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -15,6 +15,7 @@ import {
 } from '../test/testHelpers';
 
 import { EppoLocalStorage } from './local-storage';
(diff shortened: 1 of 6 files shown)


src/index.spec.ts
```diff
@@ -256,6 +257,36 @@ describe('EppoJSClient E2E test', () => {
     });
   });
 
+  describe('LocalStorageAssignmentCache', () => {
```
leoromanovsky, line 260: testing the cache directly instead of as an integration with `getAssignment` as those already existing in the commons

src/local-storage-assignment-cache.ts
```diff
@@ -0,0 +1,51 @@
+import { AssignmentCache } from '@eppo/js-client-sdk-common';
+
+import { hasWindowLocalStorage } from './local-storage';
+
+class LocalStorageAssignmentShim {
+  LOCAL_STORAGE_KEY = 'EPPO_LOCAL_STORAGE_ASSIGNMENT_CACHE';
+
+  public has(key: string): boolean {
+    if (!hasWindowLocalStorage()) {
+      return false;
```
leoromanovsky, line 10: Thought about throwing an error here however there's nothing with people disabling this API and functionally the SDK should continue to work.

sameerank: Makes sense to me too. No error throwing here 👍

sameerank (approved):
I think you'll want to increment the version too before merging to get this published

src/local-storage.ts
```diff
@@ -9,21 +9,21 @@ export class EppoLocalStorage {
...
     }
   }
 }
+
+// Checks whether local storage is enabled in the browser (the user might have disabled it).
+export function hasWindowLocalStorage(): boolean {
+  try {
+    return typeof window !== 'undefined' && !!window.localStorage;
+  } catch {
+    // Chrome throws an error if local storage is disabled and you try to access it
```
aarsilv, line 26: interesting!

aarsilv (approved):
Nice!

Outside the scope of this PR, but I wonder if we want some sort of TTL for bandits or something. Something to ponder!

PR #45, opened 2024-01-19
Non-fatal polling of request configurations via leveraging common poller

aarsilv:
🎟️  **Ticket:** [FF-1454 - Client JavaScript SDK - Non Fatal Polling](https://linear.app/eppo/issue/FF-1454/client-javascript-sdk-non-fatal-polling)

## Motivation and Context
So that our client-side JavaScript SDK can benefit from the polling improvements done to our Node SDK, we're moving `poller` to `js-client-sdk-common`. See that repository's [PR #35 - Common configuration requestor/poller ](https://github.com/Eppo-exp/js-client-sdk-common/pull/35). Now, this SDK can take advantage of it, too!

## Description
Leverages the common poller and exposes new initialization options for polling. However, one difference from the Node SDK is that there is a `pollAfterSuccessfulInitialization` option. It defaults to `false` (preserving current behavior). If set to `true`, this SDK will regularly poll for new configurations just like the Node SDK does.

## How has this been tested?
* Existing unit tests
* Added new tests to `index.spec.ts`

PR #48, opened 2024-02-03
Default uninitialized EppoClient instance

aarsilv:
🎟️ _(Eppo Internal)_ [FF-1557 - JS Client SDK should return null when not initialized](https://linear.app/eppo/issue/FF-1557/js-client-sdk-should-return-null-when-not-initialized)
👯  **Related PR:** [`js-client-sdk` #37](https://github.com/Eppo-exp/js-client-sdk-common/pull/37)

#45 introduced a change of behavior, where calls to `getInstance()` before the client was initialized would throw an error instead of returning an uninitialized client. This Pull request reverts that behavior change so that there will always be a client, and if uninitialized, it will just return `null` on any assignment calls.

This is done by instantiating a client statically at import time--as we were previously doing ([link](https://github.com/Eppo-exp/js-client-sdk/pull/45/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80L53))--as opposed to during `init()`. 

So that we can still leverage the client handling polling for configuration requests, we use `EppoClient`s newly-added (see the [related PR](https://github.com/Eppo-exp/js-client-sdk-common/pull/37))`setConfigurationRequestParameters()` where we can pass in the configuration request parameters during `init()` instead of needing the when constructing the instance.


docs/js-client-sdk.eppojsclient.initialized.md
```diff
@@ -0,0 +1,11 @@
+<!-- Do not edit this file. It is automatically generated by API Documenter. -->
+
+[Home](./index.md) &gt; [@eppo/js-client-sdk](./js-client-sdk.md) &gt; [EppoJSClient](./js-client-sdk.eppojsclient.md) &gt; [initialized](./js-client-sdk.eppojsclient.initialized.md)
+
+## EppoJSClient.initialized property
+
+**Signature:**
+
+```typescript
+static initialized: boolean;
```
aarsilv, line 10: Docs got updated because the way `index.ts` is structured, `initialized` has to be public 🙄

docs/js-client-sdk.iclientconfig.throwonfailedinitialization.md
```diff
@@ -4,7 +4,7 @@
 
 ## IClientConfig.throwOnFailedInitialization property
 
-Throw on error if unable to fetch an initial configuration during initialization. (default: true)
+Throw an error if unable to fetch an initial configuration during initialization. (default: true)
```
aarsilv, line 7: 🧹  Grammar! 📈

package.json
```diff
@@ -58,7 +58,7 @@
     "xhr-mock": "^2.5.1"
   },
   "dependencies": {
-    "@eppo/js-client-sdk-common": "2.1.0",
+    "@eppo/js-client-sdk-common": "file:../js-client-sdk-common",
```
aarsilv, line 61: This will be replaced with `"2.1.1"` once published

src/index.spec.ts
```diff
@@ -379,7 +382,7 @@ describe('EppoJSClient E2E test', () => {
       flags: {
         [hashedFlagKey]: mockExperimentConfig,
       },
-    };
+    } as unknown as Record<string, IExperimentConfiguration>;
```
aarsilv, line 385: TypeScript insisted on the double `as` because we are wiling out with our types here.

src/index.spec.ts
```diff
@@ -545,5 +548,37 @@ describe('EppoJSClient E2E test', () => {
...
+      // eslint-disable-next-line @typescript-eslint/ban-types
+      let getInstance: Function;
+      beforeEach(() => {
+        jest.isolateModules(() => {
+          // Isolate and re-require so that the static instance is reset to it's default state
+          // eslint-disable-next-line @typescript-eslint/no-var-requires
+          const reloadedModule = require('./index');
+          init = reloadedModule.init;
+          getInstance = reloadedModule.getInstance;
+        });
```
aarsilv, line 564: This `isolateModules()` and `require()` combo to reset the static EppoJSClient instance took me a bit to figure out and it had weird side effects, such as making test double (`td`) not work (hence me using `returnRac()` instead of mocking a response like other tests cases). Also, using this outside of a `beforeEach()`--which I did at first because I just have one test case--results in `console.log()` not actually logging anything 🤯 

TL;DR; use singletons and `jest.isolateModules()` with extreme caution, but it is a potentially useful tool to have your toolbox.

src/index.ts
```diff
@@ -69,12 +69,15 @@ export interface IClientConfig {
 
 export { IAssignmentLogger, IAssignmentEvent, IEppoClient } from '@eppo/js-client-sdk-common';
 
+const localStorage = new EppoLocalStorage();
```
aarsilv, line 72: Moved this back to the module-level (where it was) so it can be used in the static initialization of `EppoJSClient.instance`

src/index.ts
```diff
@@ -83,6 +86,7 @@ export class EppoJSClient extends EppoClient {
     subjectAttributes?: Record<string, any>,
     assignmentHooks?: IAssignmentHooks,
   ): string | null {
+    EppoJSClient.getAssignmentInitializationCheck();
```
aarsilv, line 89: I think we _do_ want to warn people who request assignments on uninitialized instances.

src/index.ts
```diff
@@ -214,8 +226,5 @@ export async function init(config: IClientConfig): Promise<IEppoClient> {
  * @public
  */
 export function getInstance(): IEppoClient {
-  if (!EppoJSClient.instance) {
-    throw Error('init() must first be called to initialize a client instance');
-  }
```
aarsilv, line 219: Farewell...for now... 😢

sameerank (approved):

PR #55, opened 2024-04-10
update js client to support ufc update

vpai:
Fixes: FF-1750

## Motivation and Context
Here I propogate the UFC changes made to the [common JS library](https://github.com/Eppo-exp/js-client-sdk-common/pull/45) into the JS client.

## How has this been tested?
Updated unit tests after running this repo and the updated branch side by side locally. Until that branch is merged and a new version is published, Github tests won't pass.

Reviewers, please take a look at the updated spec, that is the part of the PR I'm least sure about.

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -4,12 +4,13 @@
 
 import { createHash } from 'crypto';
 
-import { HttpClient, Flag, VariationType } from '@eppo/js-client-sdk-common';
-import { POLL_INTERVAL_MS, POLL_JITTER_PCT } from '@eppo/js-client-sdk-common/dist/constants';
+import { HttpClient, Flag, VariationType, constants } from '@eppo/js-client-sdk-common';
 import * as md5 from 'md5';
 import * as td from 'testdouble';
 import { encode } from 'universal-base64';
 import mock from 'xhr-mock';
+
+const { POLL_INTERVAL_MS, POLL_JITTER_PCT } = constants;
 
 import {
   IAssignmentTestCase,


ploomiz (approved):

Makefile
```diff
@@ -29,13 +29,13 @@ tempDir := ${testDataDir}temp/
 gitDataDir := ${tempDir}sdk-test-data/
 branchName := main
 githubRepoLink := https://github.com/Eppo-exp/sdk-test-data.git
+repoName := sdk-test-data
```
aarsilv, line 32: I don't think this variable is used anywhere

src/index.spec.ts
```diff
@@ -4,87 +4,143 @@
...
 
 import { EppoLocalStorage } from './local-storage';
 import { LocalStorageAssignmentCache } from './local-storage-assignment-cache';
 
-import { IAssignmentLogger, IEppoClient, getInstance, init } from './index';
+import { EppoJSClient, IAssignmentLogger, IEppoClient, getInstance, init } from './index';
+
+const flagEndpoint = /flag_config\/v1\/config*/;
+
+function getTestAssignments(
```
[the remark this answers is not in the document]
aarsilv: Food for possibly later thought:
We do this a lot, basically replicating the loading and then running test case logic in each SDK.

I wonder if instead, the SDK test repo has a program that runs and shells out to the command line.
Then each SDK just needs to implement a script that is takes a config, subject keys, attributes, and spits out the variations.

While we have common test _data_, this would ensure the _tests_ are all being run the same.

src/index.spec.ts
```diff
@@ -4,87 +4,143 @@
...
+    }
+    expect(subject.assignment).toEqual(assignment);
+  }
+}
 
 describe('EppoJSClient E2E test', () => {
   let globalClient: IEppoClient;
   let mockLogger: IAssignmentLogger;
-  let returnRac = readMockRacResponse; // function so it can be overridden per-test
+  let returnUfc = readMockUFCResponse; // function so it can be overridden per-test
```
aarsilv, line 75: Minor, but I'd stay consistent with camel casing our acronym as well. `readMockUfcResponse`

src/index.spec.ts
```diff
@@ -104,49 +163,10 @@ describe('EppoJSClient E2E test', () => {
     mock.teardown();
   });
 
-  it('assigns subject from overrides when experiment is enabled', () => {
```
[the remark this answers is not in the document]
aarsilv: 🪓

src/index.spec.ts
```diff
@@ -104,173 +160,183 @@ describe('EppoJSClient E2E test', () => {
...
-    const assignment = globalClient.getAssignment('subject-10', flagKey);
-    expect(assignment).toEqual(null);
+    const assignment = globalClient.getStringAssignment('subject-10', flagKey, '');
+    expect(assignment).toEqual('');
   });
 
   it('logs variation assignment and experiment key', () => {
     td.replace(EppoLocalStorage.prototype, 'get', (key: string) => {
-      if (key !== hashedFlagKey) {
+      if (key !== flagKey) {
```
aarsilv, line 172: Do we ever expect the client-side JS client to ever _not_ use hashed/obfuscated configurations?
I could see ituseful for local debugging.

If so, I wonder if, in our top-level UFC object, we include `obfuscated=true/false` as a hint to the SDK if it's at play or not.

src/index.spec.ts
```diff
@@ -104,173 +160,183 @@ describe('EppoJSClient E2E test', () => {
...
-        const assignments = getAssignmentsWithSubjectAttributes(
-          subjectsWithAttributes
-            ? subjectsWithAttributes
-            : subjects.map((subject) => ({ subjectKey: subject })),
-          experiment,
-          valueType,
+      async ({ flag, variationType, defaultValue, subjects }: IAssignmentTestCase) => {
+        const client = new EppoJSClient(new EppoLocalStorage());
+
+        let assignments: {
```
aarsilv, line 264: Any reasons we're declaring this here just to overwrite it later?

src/index.spec.ts
```diff
@@ -104,173 +160,183 @@ describe('EppoJSClient E2E test', () => {
...
+          [VariationType.NUMERIC]: client.getNumericAssignment.bind(client),
+          [VariationType.INTEGER]: client.getIntegerAssignment.bind(client),
+          [VariationType.STRING]: client.getStringAssignment.bind(client),
+          [VariationType.JSON]: client.getJSONAssignment.bind(client),
+        };
+
+        const assignmentFn = typeAssignmentFunctions[variationType];
+        if (!assignmentFn) {
+          throw new Error(`Unknown variation type: ${variationType}`);
+        }
```
aarsilv, line 280: ❤️ 
I too prefer this method vs the cumbersome `switch` statement

src/index.spec.ts
```diff
@@ -104,173 +160,183 @@ describe('EppoJSClient E2E test', () => {
...
-            expect(jsonStringAssignments).toEqual(expectedAssignments);
-            break;
-          }
-        }
+        validateTestAssignments(assignments, flag);
       },
     );
+  });
+
+  describe('UFC Obfuscated Test Cases', () => {
```
aarsilv, line 293: 🔥

src/index.spec.ts
```diff
@@ -104,173 +160,183 @@ describe('EppoJSClient E2E test', () => {
...
-    expect(assignment).toEqual('control');
+    expect(assignment).toEqual('variant-1');
   });
 
-  describe('getAssignment', () => {
-    const testData = readAssignmentTestData();
-
-    it.each(testData)(
+  describe('UFC General Test Cases', () => {
+    it.each(readAssignmentTestData())(
```
aarsilv, line 259: This is the basically the same as the other method with the exception of which config we're using in our moc. I bet we could fold them into the same with something like:
```ts
const unobfuscatedTests = readAssignmentTestData();
const obfuscatedTests = unobfuscatedTests.map((t) => ({obfuscate: true, ...t}));
it.each(unobfuscatedTests.concat(obfuscatedTests))(
```
Then in, the below code, if its obfuscated, we do the obfuscated mocking/init

src/index.spec.ts
```diff
@@ -565,19 +575,19 @@ describe('EppoJSClient E2E test', () => {
...
           apiKey,
           baseUrl,
           assignmentLogger: mockLogger,
         });
-        expect(client.getStringAssignment('subject', flagKey)).toBeNull();
+        expect(client.getStringAssignment('subject', flagKey, '')).toBe('');
         // Advance time so a poll happened and check again
         await jest.advanceTimersByTimeAsync(POLL_INTERVAL_MS);
-        expect(client.getStringAssignment('subject', flagKey)).toBe('control');
+        expect(client.getStringAssignment('subject', flagKey, '')).toBe('control');
```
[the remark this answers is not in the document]
aarsilv: 😞 
Yeah happy to pair!

src/index.ts
```diff
@@ -79,88 +78,54 @@ export class EppoJSClient extends EppoClient {
...
-    subjectKey: string,
-    flagKey: string,
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    subjectAttributes?: Record<string, any>,
-    assignmentHooks?: IAssignmentHooks,
-  ): string | null {
-    EppoJSClient.getAssignmentInitializationCheck();
-    return super.getAssignment(subjectKey, flagKey, subjectAttributes, assignmentHooks, true);
-  }
-
```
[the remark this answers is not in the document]
aarsilv: 🪓

src/local-storage.spec.ts
```diff
@@ -43,5 +43,10 @@ describe('EppoLocalStorage', () => {
       expect(storage.get<ITestEntry>('key1')).toEqual(config1);
       expect(storage.get<ITestEntry>('key2')).toEqual(config2);
     });
+
+    it('returns a list of keys', () => {
+      storage.setEntries({ key1: config1, key2: config2 });
+      expect(storage.getKeys()).toEqual(['key1', 'key2']);
+    });
```
aarsilv, line 50: 📈  Tests save lives!

src/local-storage.ts
```diff
@@ -17,6 +21,13 @@ export class EppoLocalStorage {
     return null;
   }
 
+  public getKeys(): string[] {
+    if (hasWindowLocalStorage()) {
+      return Object.keys(window.localStorage);
```
aarsilv, line 26: Is this gonna get EVERYTHING in local storage? I wonder if we should isolate these to all be under some Eppo local storage property.

test/testHelpers.ts
```diff
@@ -1,11 +1,13 @@
 import * as fs from 'fs';
```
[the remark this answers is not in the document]
aarsilv: If it's exported, could we just import these methods from the common repo and drop the duplicate file entirely?

aarsilv (approved):
Looking good to me! Happy to pair on the (curiously) flakey test!

src/index.spec.ts
```diff
@@ -4,87 +4,143 @@
...
-            value: 'variant-2',
-            typedValue: 'variant-2',
-            shardRange: {
-              start: 67,
-              end: 100,
-            },
+            variationKey: 'variant-2',
+            shards: [
+              {
+                salt: 'some-sat',
```
schmit, line 127:
```suggestion
                salt: 'some-salt',
```

src/index.spec.ts
```diff
@@ -104,173 +160,183 @@ describe('EppoJSClient E2E test', () => {
...
-  });
-
-  it('returns null when experiment config is absent', () => {
+  it('returns default value when experiment config is absent', () => {
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     td.replace(EppoLocalStorage.prototype, 'get', (key: string) => null as null);
-    const assignment = globalClient.getAssignment('subject-10', flagKey);
-    expect(assignment).toEqual(null);
+    const assignment = globalClient.getStringAssignment('subject-10', flagKey, '');
+    expect(assignment).toEqual('');
```
schmit, line 167: nit: think this makes it a little more obvious what's happening

```suggestion
    const assignment = globalClient.getStringAssignment('subject-10', flagKey, 'default-value');
    expect(assignment).toEqual('default-value');
```

schmit (approved):
LGTM! Added a few minor comments. Note I am updating the test suite with some extra tests: https://github.com/Eppo-exp/sdk-test-data/pull/14, let's make sure these also pass once merged

docs/js-client-sdk.eppojsclient.getboolassignment.md
```diff
@@ -7,7 +7,7 @@
 **Signature:**
 
 ```typescript
-getBoolAssignment(subjectKey: string, flagKey: string, subjectAttributes?: Record<string, any>, assignmentHooks?: IAssignmentHooks): boolean | null;
+getBoolAssignment(subjectKey: string, flagKey: string, defaultValue: boolean, subjectAttributes?: Record<string, any>): boolean | null;
```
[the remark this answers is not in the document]
giorgiomartini0: I just realized this file is autogenerated... so the problem is upstream

js-client-sdk.api.md
```diff
@@ -6,24 +6,21 @@
...
 import { IAssignmentEvent } from '@eppo/js-client-sdk-common';
-import { IAssignmentHooks } from '@eppo/js-client-sdk-common';
 import { IAssignmentLogger } from '@eppo/js-client-sdk-common';
 import { IEppoClient } from '@eppo/js-client-sdk-common';
 
 // @public
 export class EppoJSClient extends EppoClient {
     // (undocumented)
-    getAssignment(subjectKey: string, flagKey: string, subjectAttributes?: Record<string, any>, assignmentHooks?: IAssignmentHooks): string | null;
+    getBoolAssignment(subjectKey: string, flagKey: string, defaultValue: boolean, subjectAttributes?: Record<string, any>): boolean | null;
```
giorgiomartini0, line 15: Another autogenerated file with the wrong return type (should not be `| null`).

src/index.spec.ts
```diff
@@ -4,87 +4,143 @@
...
+    );
+    assignments.push({ subject: subject, assignment: assignment });
+  }
+  return assignments;
+}
+
+function validateTestAssignments(
+  assignments: {
+    subject: SubjectTestCase;
+    assignment: string | boolean | number | object | null;
```
giorgiomartini0, line 53: `null` should not be allowed here.

src/index.spec.ts
```diff
@@ -104,173 +160,183 @@ describe('EppoJSClient E2E test', () => {
...
     });
 
     const subjectAttributes = { foo: 3 };
     globalClient.setLogger(mockLogger);
-    const assignment = globalClient.getAssignment('subject-10', flagKey, subjectAttributes);
-    expect(assignment).toEqual('control');
+    const assignment = globalClient.getStringAssignment(
+      'subject-10',
+      flagKey,
+      '',
```
giorgiomartini0, line 184: Similar to Sven's comment above, I'd prefer us showcasing informative defaults.

src/index.ts
```diff
@@ -79,88 +78,54 @@ export class EppoJSClient extends EppoClient {
...
   public getStringAssignment(
     subjectKey: string,
     flagKey: string,
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    defaultValue: string,
     subjectAttributes?: Record<string, any>,
-    assignmentHooks?: IAssignmentHooks,
   ): string | null {
```
giorgiomartini0, line 86: This should not have `| null`, and similarly for the other getters.

giorgiomartini0 (changes requested):
Only blocking request: remove `| null` from the return signature of all the get<Type>Assignment. This should then fix the API docs automatically.

src/index.ts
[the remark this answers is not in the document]
ploomiz: I will miss it.

src/index.spec.ts
giorgiomartini0: @vpai Not sure if you missed this one or if the common repo also needs to be updated. Everything else looks good!

giorgiomartini0 (approved):

PR #64, opened 2024-05-15
Add support for skipInitialPoll parameter (FF-2084)

leoromanovsky:
---
labels: mergeable
---
[//]: #  (Link to the issue corresponding to this chunk of work)
Fixes: #__issue__

## Motivation and Context
[//]: #  (Why is this change required? What problem does it solve?)

Customers can elect to put the client SDK into "offline" mode - one of the configurations they need to enable is preventing the SDK from performing a CDN fetch during the `init` method.

Subsequent calls to the CDN will be blocked by setting the `isExpired` function in the persistence store implementations.

## Description
[//]: # (Describe your changes in detail)

* adds support for js-commons `skipInitialPoll`

## How has this been tested?
[//]: # (Please describe in detail how you tested your changes)

[//]: # (OPTIONAL)
[//]: # (Add one or multiple labels: enhancement, refactoring, bugfix)

PR #65, opened 2024-05-15
Provide a customer persistence store (FF-2085)

leoromanovsky:
---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context

Allow customers to provide a customer persistence store implementation of `IAsyncStore`. They can pair this with (https://github.com/Eppo-exp/js-client-sdk/pull/64) to create an "off-line" mode where the client CDN will not make any remote requests. In this case they would provide a `IAsyncStore` that reads from a local file (or memory) for a configuration that was provided by their server.

Another use case is for AWS Lambda functions, not in scope for this PR as that is nodejs, which are very short-lived servers. Their persistence store can be backed by DynamoDB to avoid many CDN requests.

## Description

* adds configuration parameter for `persistenceStore` - if provided a `HybridConfigurationStore` will be created with it

**questions for reviewers**

In order to provide the persistence store, I needed to instantiate the Client singleton within the `init` method. Does this seem like an acceptable compromise? The alternative is adding a `usePersistenceStore()` method onto js-commons.

## How has this been tested?


aarsilv (approved):
❤️

sameerank (approved):

src/configuration-factory.ts
```diff
@@ -1,15 +1,20 @@
...
   MemoryOnlyConfigurationStore,
   MemoryStore,
 } from '@eppo/js-client-sdk-common';
 
 import { LocalStorageBackedAsyncStore } from './local-storage';
 
-export function configurationStorageFactory(): IConfigurationStore<Flag> {
-  if (hasWindowLocalStorage()) {
+export function configurationStorageFactory(
+  persistenceStore?: IAsyncStore<Flag>,
```
sameerank, line 13: Consider renaming to `persistentStore`, which I think is what this is usually called, unless I'm misunderstanding what this is

sameerank: And I think it makes sense as an argument in init rather than a `usePersistentStore(store)` method that could modify the client's storage after it has been initialized. If there's a scenario where swapping stores is useful, it doesn't seem like we're blocked from providing that method in the future.

leoromanovsky: > And I think it makes sense as an argument in init rather than a usePersistentStore(store) method that could modify the client's storage after it has been initialized

That was my thinking too - can't imagine wanting to swap this out in the middle of a session.

PR #69, opened 2024-05-17
initialize the configuration store

leoromanovsky:
---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context

1️⃣  Omitting calling the `init` method on the `IConfigurationStore` means that the `servingStore` is not populated from whatever is present on disk (localStorage, chrome.storage, etc).

Throttling the browser network demonstrates that the flags don't begin evaluating to `true` until after the network request completes, whereas the desired behavior is already locally cached flags should start being evaluated as soon as the disk cache is loaded into memory, before the network request:

https://github.com/Eppo-exp/js-client-sdk/assets/57361/ac3a610c-02c0-4709-8555-ab3427988286

2️⃣ If the persistent store is `null`, which should not have happen if the type checker was strict, the `SyncStore.setEntries` method fails since its implemented to assume a not-null parameter.

👉 This only happen in integration tests but was a useful insight!

## Description

* initialize the configuration store, loading the disk cache into memory to begin serving before the network request
* enable strict null checking and conform the return types

## Deployment

Publish a new RC build and deploy it to prod eppo.cloud for dogfooding.

## How has this been tested?

created an RC release of `3.0.2-rc3` and tested by deploying eppo to qa:

https://github.com/Eppo-exp/js-client-sdk/assets/57361/8a12b18b-227e-4eda-86ec-ac3ec2c0cdcf

[//]: # (OPTIONAL)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -170,6 +170,7 @@ export async function init(config: IClientConfig): Promise<IEppoClient> {
     // Set the configuration store to the desired persistent store, if provided.
     // Otherwise the factory method will detect the current environment and instantiate the correct store.
     const configurationStore = configurationStorageFactory(config.persistentStore, false);
+    await configurationStore.init();
     EppoJSClient.instance.setConfigurationStore(configurationStore);
 
     const requestConfiguration: FlagConfigurationRequestParameters = {
--- a/src/index.ts
+++ b/src/index.ts
@@ -167,10 +167,16 @@ export async function init(config: IClientConfig): Promise<IEppoClient> {
       EppoJSClient.instance.stopPolling();
     }
 
+    console.log('Initializing Eppo SDK');
+
     // Set the configuration store to the desired persistent store, if provided.
     // Otherwise the factory method will detect the current environment and instantiate the correct store.
     const configurationStore = configurationStorageFactory(config.persistentStore, false);
+    console.log('Using configuration store:', configurationStore);
+
+    console.log('(start) Initializing configuration store');
     await configurationStore.init();
+    console.log('(done) Initialized configuration store');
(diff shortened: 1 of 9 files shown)


src/local-storage.ts
```diff
@@ -22,10 +22,10 @@ export class LocalStorageBackedAsyncStore<T> implements IAsyncStore<T> {
     return Promise.resolve(true);
   }
 
-  getEntries(): Promise<Record<string, T> | null> {
```
leoromanovsky, line 25: compiler should not allowed this type to have deviated from the base class

src/index.ts
```diff
@@ -170,6 +170,7 @@ export async function init(config: IClientConfig): Promise<IEppoClient> {
     // Set the configuration store to the desired persistent store, if provided.
     // Otherwise the factory method will detect the current environment and instantiate the correct store.
     const configurationStore = configurationStorageFactory(config.persistentStore, false);
+    await configurationStore.init();
```
leoromanovsky, line 173: populates the serving store from the persistent store and makes it ready for evaluations

src/local-storage-assignment-cache.ts
```diff
@@ -13,12 +13,12 @@ class LocalStorageAssignmentShim {
...
 
-  public get(key: string): string {
+  public get(key: string): string | undefined {
     if (!hasWindowLocalStorage()) {
-      return null;
+      return undefined;
     }
 
-    return this.getCache().get(key);
+    return this.getCache().get(key) || undefined;
```
sameerank, line 21: Is a nullish (`??`) more appropriate for not finding a matching key?

sameerank (approved):

schmit (approved):
Good catch and strict null checks 🙌

PR #71, opened 2024-05-19
If localStorage is not available, use a in-memory none expiring assig…

leoromanovsky:
…nment cache (FF-2110)

---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context

When `init` is called, a `localStorage` assignment cache is created. However if no `localStorage` is available it always returns non-cachable return values, effectively being no cache. This is going to happen on chrome extensions and other environments without a `localStorage`.

## Description

* If no `localStorage` is available, create an in-memory assignment cache. This is an improvement over the current behavior as at least there will be some caching happening, even though it is reset each session.
* extracted helper methods in `environment.ts`
* moved assignment cache files into own directory

## How has this been tested?

[//]: # (OPTIONAL)

### DIFF
--- a/src/assignment-cache/local-storage-assignment-cache.ts
+++ b/src/assignment-cache/local-storage-assignment-cache.ts
@@ -1,31 +1,17 @@
 import { AssignmentCache } from '@eppo/js-client-sdk-common';
 
-import { hasWindowLocalStorage } from './configuration-factory';
-
 class LocalStorageAssignmentShim {
   LOCAL_STORAGE_KEY = 'EPPO_LOCAL_STORAGE_ASSIGNMENT_CACHE';
 
   public has(key: string): boolean {
-    if (!hasWindowLocalStorage()) {
-      return false;
-    }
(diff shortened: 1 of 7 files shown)


src/assignment-cache/local-storage-assignment-cache.ts
```diff
@@ -1,31 +1,17 @@
...
 
-import { hasWindowLocalStorage } from './configuration-factory';
-
 class LocalStorageAssignmentShim {
   LOCAL_STORAGE_KEY = 'EPPO_LOCAL_STORAGE_ASSIGNMENT_CACHE';
 
   public has(key: string): boolean {
-    if (!hasWindowLocalStorage()) {
-      return false;
-    }
```
leoromanovsky, line 11: IMO no need for these methods - the class should only be instantiated from the factory and when `localStorage` is detected. If not, throwing an exception is "good" behavior so the end-user uses the more appropriate class.

PR #73, opened 2024-05-28
Configurable cache behavior

aarsilv:
_Eppo Internal:_
🎟️ **Ticket:** [FF-2218 - JS SDK - Ensure desired initialization cache and fetch outcomes](https://linear.app/eppo/issue/FF-2218/js-sdk-ensure-desired-initialization-cache-and-fetch-outcomes)

This PR makes it so that if the config loads a non-zero amount of flags from the cache, it doesn't wait for fetching an updated configuration to complete before resolving the initialization promise. In the unexpected case fetch finishes first, the cache will be ignored (but still updated). If one fails but not the other, then init will successfully resolve once the first method succeeds. If both fail, it will reject.


src/index.ts
```diff
@@ -207,25 +211,79 @@ export async function init(config: IClientConfig): Promise<IEppoClient> {
...
+      })
+      .catch((e) => {
+        console.warn('Eppo SDK encountered an error initializing from fetching', e);
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        initFromFetchError = e;
+      });
+
+    console.log('>>>>> ABOUT TO RACE!');
+
+    let initializationSource = await Promise.race([
```
aarsilv, line 247: [`Promise.race()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) will continue execution as soon as the first promise resolves (successfully or unsuccessfully)

This let's us (potentially) resolve initialization as soon as the cache or fetch finishes without waiting for the other.

src/index.ts
```diff
@@ -207,25 +211,79 @@ export async function init(config: IClientConfig): Promise<IEppoClient> {
...
+      attemptInitFromConfigStore,
+      attemptInitFromFetch,
+    ]);
+
+    console.log('>>>> loaded', configurationStore.getKeys().length, 'flags from', initializationSource, configurationStore.getKeys());
+
+    if (!initializationSource) {
```
aarsilv, line 252: because of the `catch()` statements in the racing promises, no error will be _thrown_ upon race completion, but if the "winner" failed then no source will be set from that method, and we need to try the other. We can tell because we capture errors.

package.json
```diff
@@ -58,7 +57,6 @@
     "webpack-cli": "^4.10.0"
   },
   "dependencies": {
-    "@eppo/js-client-sdk-common": "3.0.8",
-    "md5": "^2.3.0"
```
aarsilv, line 62: we do not need an md5 library in this package

src/index.spec.ts
```diff
@@ -354,7 +347,8 @@ describe('initialization options', () => {
       ],
     });
 
-    // We're creating a new instance for each test so we need to clear the underlying storage too
+    // We want each test to have an empty local storage
+    // Note: mock localStorage is provided by jest
```
aarsilv, line 351: Took me a while to figure this out 😅

src/index.ts
```diff
@@ -110,7 +111,7 @@ export class EppoJSClient extends EppoClient {
   public getStringAssignment(
     flagKey: string,
     subjectKey: string,
-    subjectAttributes: Record<string, any>,
+    subjectAttributes: Record<string, AttributeType>,
```
aarsilv, line 114: This is the signature the common repo function expects

yarn.lock
```diff
@@ -1013,11 +1013,6 @@
   resolved "https://registry.yarnpkg.com/@types/json5/-/json5-0.0.29.tgz#ee28707ae94e11d2b827bcbe5270bcea7f3e71ee"
   integrity sha512-dRLjCWHYg4oaA77cxO64oO+7JwCwnIzkZPdrrC71jQmQtlhM556pwKo5bUzqvZndkVbeFLIIi+9TC40JNF5hNQ==
 
-"@types/md5@^2.3.2":
```
aarsilv, line 1016: 🪓

src/index.ts
```diff
@@ -207,25 +212,73 @@ export async function init(config: IClientConfig): Promise<IEppoClient> {
...
+          return '';
+        }
+        return 'config store';
+      })
+      .catch((e) => {
+        console.warn('Eppo SDK encountered an error initializing from the configuration store', e);
+        initFromConfigStoreError = e;
+      });
+    const attemptInitFromFetch = EppoJSClient.instance
+      .fetchFlagConfigurations()
```
leoromanovsky, line 238: Does `skipInitialPoll` continue to work as before - I assume if it is `true` then `fetchFlagConfigurations` returns.

leoromanovsky (changes requested):
Could you back up a bit and explain the need for this PR - is it focused on shortening the time for SDK initialization or is there a bug you think arises from loading the configuration from disk followed by the network request serially?

I'm mentioning as the code is well tested but adds some complexity.

src/index.ts
aarsilv: Yeah if that is true then `fetchClagConfigurations()`'s call of `await this.requestPoller.start();` will complete without the initial poll attempt firing off.

src/index.ts
```diff
@@ -207,25 +212,73 @@ export async function init(config: IClientConfig): Promise<IEppoClient> {
...
+    // We have two at-bats for initialization: from the configuration store and from fetching
+    // We can resolve the initialization promise as soon as either one succeeds
+    let initFromConfigStoreError = undefined;
+    let initFromFetchError = undefined;
 
-    // default behavior is to use a LocalStorage-based assignment cache.
-    // this can be overridden after initialization.
-    EppoJSClient.instance.useCustomAssignmentCache(new LocalStorageAssignmentCache());
-    EppoJSClient.instance.setConfigurationRequestParameters(requestConfiguration);
-    await EppoJSClient.instance.fetchFlagConfigurations();
```
aarsilv, line 217: Doing this here, "serially" with `await configurationStore.init()`, has two consequences:
1. If the cached configuration is successfully loaded, we still need to then wait for the fetch of the updated configuration before `init()` promise resolves and `initialized` is set to `true`.
2. If either loading from the cache or fetching hits an error, we will say we failed to initialize (and possibly rethrow) despite a configuration being successfully loaded

src/index.spec.ts
```diff
@@ -317,24 +320,10 @@ describe('initialization options', () => {
...
-    global.fetch = jest.fn(() => {
-      const ufc = returnUfc(MOCK_UFC_RESPONSE_FILE);
-
-      return Promise.resolve({
-        ok: true,
-        status: 200,
-        json: () => Promise.resolve(ufc),
-      });
-    }) as jest.Mock;
+    global.fetch = jest.fn(() => Promise.reject('Each test case should mock fetch'));
```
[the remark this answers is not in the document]
greghuels: Great idea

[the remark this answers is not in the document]
greghuels: It might be better to use `spyFn = jest.spyOn(global, 'fetch')` in this case so that you can restore the original function via `spy.mockRestore()` in an `afterAll`

src/index.spec.ts
```diff
@@ -528,6 +524,211 @@ describe('initialization options', () => {
     expect(callCount).toBe(0);
   });
 
+  it('Uses its cache', async () => {
+    // Mock fetch so first call works, second fails
+    let fetchCallCount = 0;
+    let fetchResolveCount = 0;
+    const fetchResolveDelayMs = 500;
+    global.fetch = jest.fn(() => {
```
greghuels, line 532: It might be better to use spyFn = jest.spyOn(global, 'fetch') in this case so that you can restore the original function via spy.mockRestore() in an afterAll

greghuels (approved):

sameerank (approved):
I think this approach using `Promise.race` is really neat!

I have some of the same concern as Leo about adding complexity without knowing if this solves any existing issues, but also having this edge case that could lead to initialization failure handled feels like an overall improvement so I approve 👍

docs/js-client-sdk.eppojsclient.getboolassignment.md
```diff
@@ -7,7 +7,7 @@
 **Signature:**
 
 ```typescript
-getBoolAssignment(flagKey: string, subjectKey: string, subjectAttributes: Record<string, any>, defaultValue: boolean): boolean;
+getBoolAssignment(flagKey: string, subjectKey: string, subjectAttributes: Record<string, AttributeType>, defaultValue: boolean): boolean;
```
aarsilv, line 10: All the autogenerated docs updated with the type change 🙄

src/abstract-string-valued.store.ts
```diff
@@ -0,0 +1,48 @@
+import { IAsyncStore } from '@eppo/js-client-sdk-common';
+
+export abstract class AbstractStringValuedAsyncStore<T> implements IAsyncStore<T> {
```
aarsilv, line 3: After adding the concept of an an "age" to the local-storage backed store, it was _very_ similar to the chrome storage backed store. So I consolidated all the logic into a shared parent abstract class.

src/chrome-storage.store.spec.ts
```diff
@@ -1,6 +1,6 @@
-import { ChromeStorageAsyncStore } from './chrome.configuration-store';
+import { ChromeStorageAsyncStore } from './chrome-storage.store';
 
-describe('ChromeStore', () => {
+describe('ChromeConfigurationStore', () => {
```
aarsilv, line 3: Tried to get the file names consistent

src/chrome-storage.store.ts
```diff
@@ -0,0 +1,31 @@
+import { AbstractStringValuedAsyncStore } from './abstract-string-valued.store';
+
+export class ChromeStorageAsyncStore<T> extends AbstractStringValuedAsyncStore<T> {
```
aarsilv, line 3: With the rename and big changes, GitHub diff thinks this is effectively a new file 🤷‍♂️ 

Compare to the "Deleted" file below

src/configuration-factory.ts
```diff
@@ -30,18 +37,24 @@ export function configurationStorageFactory(
   if (forceMemoryOnly) {
     return new MemoryOnlyConfigurationStore();
   } else if (persistentStore) {
-    return new HybridConfigurationStore(new MemoryStore<Flag>(), persistentStore);
+    return new IsolatedHybridConfigurationStore(
```
aarsilv, line 40: Introduced an `IsolatedHybridConfigurationStore` to allow isolating updates to the serving store from updates of the persistent store--useful in situations where we want per-session consistency.

src/configuration-factory.ts
```diff
@@ -30,18 +37,24 @@ export function configurationStorageFactory(
...
+      new MemoryStore<Flag>(),
+      persistentStore,
+      servingStoreUpdateStrategy,
```
aarsilv, line 43: This "update strategy" tells it how to isolate the stores. The default of "always" makes it behave just like the non-isolated `HybridStore` that updates both stores every call to `setEntries()`.

src/index.spec.ts
aarsilv: great call!

src/index.spec.ts
```diff
@@ -528,6 +523,207 @@ describe('initialization options', () => {
...
+    await jest.advanceTimersByTimeAsync(mockStoreDelayMs);
+    expect(fetchCallCount).toBe(1);
+    expect(storeLoaded).toBe(true);
+    expect(client.getStringAssignment(flagKey, 'subject', {}, 'default-value')).toBe('control');
+    // store entries should reflect latest fetch
+    expect(mockStoreEntries).toEqual(mockConfigResponse.flags);
+  });
+
+  test.each(['always', 'expired', 'empty'])(
+    'Fetch updates cache according to the "%s" update strategy',
```
aarsilv, line 638: Newly-added test for how the optional "update strategy" configuration option plays out during init

src/local-storage.store.spec.ts
```diff
@@ -0,0 +1,59 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import { LocalStorageBackedAsyncStore } from './local-storage.store';
+
+describe('LocalStorageBackedAsyncStore', () => {
```
aarsilv, line 7: Renaming plus extending makes GitHub diff thing this is a new file 🤷‍♂️

src/local-storage.store.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/*
+ * Removing stale keys is facilitating by storing the entire configuration into a single
+ * key in local storage. This is done by serializing the entire configuration object
+ * into a string and then storing it to a single key. When retrieving the configuration,
+ * we first check if the configuration exists in local storage. If it does, we deserialize
+ * the configuration from the string and return it. If the configuration does not exist,
+ * we return null.
+ */
+export class LocalStorageBackedAsyncStore<T> extends AbstractStringValuedAsyncStore<T> {
```
aarsilv, line 11: Renaming plus extending makes GitHub diff thing this is a new file as well 🤷‍♂️

src/local-storage.store.ts
```diff
@@ -0,0 +1,31 @@
...
+  constructor(private localStorage: Storage, cooldownSeconds?: number) {
```
aarsilv, line 12: added optoinal `cooldownSeconds?:` like we have for the chrome storage, that way the client has knowledge of the age of the data loaded from local storage, and can act appropriately.

aarsilv (commented):
I've expanded this PR to include two new configuration options for the client with defaults that mirror current behavior:
* `maxCacheAgeSeconds`: Maximum age, in seconds, previously cached values are considered valid until new values will be fetched (default: 0)
* `updateOnFetch`: Sets how the configuration is updated after a successful fetch
  * always: immediately start using the new configuration
  * expired: immediately start using the new configuration only if the current one has expired
  * empty: only use the new configuration if the current one is both expired and uninitialized/empty
  
To support different update options, I also introduced an `IsolatedHybridConfigurationStore` that doesn't necessarily always update the serving store. (For example, the user may want to maintain consistency within a session)

Remaining for this PR is factoring in a `useExpiredCache` option, which will help control when the initialization finishes `await`ing. This option plus the two above achieve the flexibility as outlined in [this Notion document](https://www.notion.so/eppo/Client-SDK-Caching-and-Fetching-Behavior-a9c8caf98c7e4a27b3d4c7751ad65cd9?pvs=4). It will be valuable to see if/how customers use it to help us decide if we want this to become standard for all client SDKs.

src/index.spec.ts
aarsilv: Alright we need to do this after all because jest doesn't allow `fetch` in global by default; probably because node didn't have it until version 18 🤷‍♂️ 

https://github.com/jestjs/jest/issues/2071

Doing it the "correct" way results in an error 😞

src/index.ts
```diff
@@ -71,17 +73,38 @@ export interface IClientConfig {
    */
   numPollRequestRetries?: number;
 
+  /**
+   * Skip the request for new configurations during initialization. (default: false)
+   */
+  skipInitialRequest?: boolean;
```
aarsilv, line 79: I didn't add this one, just moved it up here.

aarsilv:
@leoromanovsky, @sameerank, & @greghuels 

Sorry to do the awful "request" of a review, but this PR has changed substantially since you last reviewed it. Mainly, instead of forcing our choice of speed over consistency on the customer, we allow them to configure things as they seem fit. It will be interesting to see if/how they use these things.

src/isoldated-hybrid.store.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    getEntries: jest.fn(),
+    isInitialized: jest.fn(),
+    isExpired: jest.fn(),
+    setEntries: jest.fn(),
+  };
+
+  const servingStoreUpdateStrategies: ServingStoreUpdateStrategy[] = ['always', 'expired', 'empty'];
+
+  // Dynamically create a describe() block for stores initialized with each of the three update strategies
+  servingStoreUpdateStrategies.forEach((updateStrategy) => {
```
greghuels, line 24: Non-blocking/nit: I've found wrapping tests in `forEach` blocks can make it difficult to focus on a test if something is failing. While not an issue here, it's also easy to forget to make sure the test names are differentiated via string interpolation. I generally prefer a little repetition with helper functions to centralize code. (none of that is really that big of a deal though, so there's really no need to spend time to make changes)

greghuels: Oh yeah, jest also has `describe.each` and `it.each` built in to it.

greghuels (approved):
Main feedback is using a composition approach over inheritance, but nothing blocking.

src/index.ts
greghuels: Or is this to prevent some misuse of init where a customer is spamming us needlessly?

src/abstract-string-valued.store.ts
aarsilv: I like this idea; thanks for putting together an example, too! The old adage is to favor composition over inheritance!

I think I went down this path because of the "is a" relationship. Chrome Storage is a String-valued store. Local Storage is a String-valued store. But it can also be interpreted as "has a"-- A string-valued store has a chrome storage engine, or it has a local storage engine. (Whereas an animal cannot have a dog). Since that second relationship _also_ makes sense, I say the tie goes to composition!

src/index.ts
aarsilv: Yeah a bit of history here, we used to just pass in a (synchronous) configuration store to the common client. For the client SDK, this was backed by local storage. There are some configuration options in the common SDK regarding what to do if the initial request fails. This let clients do things like fall back on serving whatever was in local storage. Some customers would `await` and others wouldn't. This also let them serve stale stuff until fresh stuff came in.

However, two things changed:
1. We have customers who want _asynchronous_ persistent stores (thus the introduction of the hybrid store)
2. We have customers who want finer control over when fetches happen, when cached data is used, etc. (thus this PR)

Some customers care about consistency. Others--such as retailers where there is a direct line from seconds of loading time to revenue--care about speed. The recent work had us biasing towards the former while other customers (e.g., clerk) were asking about the latter. One of the consistency concerns is avoiding layout shift, this mean a customer may want to roll with an expired configuration and just use the updated one next time.

src/isoldated-hybrid.store.spec.ts
aarsilv: I know about `it.each()` but didn't know about `describe.each()` -- I'll use that!

Agree on your concern that it can be hard to know which test is failing but I think putting the interpolation in is worth the simplicity of using the data provider.

.gitignore
```diff
@@ -10,3 +10,5 @@ yarn-error.log
 test/data
 
 .eslintcache
+
+.idea/
```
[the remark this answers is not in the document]
felipecsl: yes :)

src/index.spec.ts
greghuels: Ah ok

src/abstract-string-valued.store.ts
felipecsl: would recommend leaving top level class comments, this looks like a good candidate!

src/isoldated-hybrid.store.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+
+      it('should initialize the serving store with entries from the persistent store', async () => {
+        const entries = { key1: 'value1', key2: 'value2' };
+        asyncStoreMock.isInitialized.mockReturnValue(true);
+        asyncStoreMock.getEntries.mockResolvedValue(entries);
+
+        await store.init();
+
+        expect(syncStoreMock.setEntries).toHaveBeenCalledWith(entries);
+        expect(asyncStoreMock.setEntries).not.toHaveBeenCalled();
```
felipecsl, line 44: sorry if this is not very constructive feedback, but I'd generally consider a "code smell" when there's heavy use of mocking in tests, especially around internal state of a class. You're basically making a lot of assumptions about how the internal state of `asyncStoreMock` should look like, especially mocking `isInitialized` can be problematic. I'd prefer using fake implementations instead. this is a non-blocking comment, just a general observation

src/chrome-storage.store.ts
```diff
@@ -0,0 +1,31 @@
...
+  private chromeStorageKey = 'eppo-sdk';
```
felipecsl, line 4: `readonly`

src/isoldated-hybrid.store.spec.ts
aarsilv: Ah yes I agree in general, what is happening here is we are basically making a dummy store / fake implementation using Jest's mocking framework. I _think_ this approach saves some effort over making a full fake, as we only need implementations for functions we care about but agree its in a gray area for sure.

src/isoldated-hybrid.store.spec.ts
```diff
@@ -0,0 +1,132 @@
+import {
```
leoromanovsky, line 1: type in this filename

src/index.spec.ts
```diff
@@ -528,6 +523,285 @@ describe('initialization options', () => {
...
+    });
+
+    // Should serve assignment from cache before fetch even fails
+    expect(fetchCallCount).toBe(2);
+    expect(fetchResolveCount).toBe(1);
+    expect(client.getStringAssignment(flagKey, 'subject', {}, 'default-value')).toBe('control');
+
+    // Should not be tripped up by failed fetch
+    await jest.advanceTimersByTimeAsync(fetchResolveDelayMs);
+    expect(fetchResolveCount).toBe(2);
```
leoromanovsky, line 647: great tests

src/index.spec.ts
```diff
@@ -528,6 +523,285 @@ describe('initialization options', () => {
...
+        'default-value',
+      );
+
+      // Completed fetch will now result in updated assignments unless set not to update a previously
+      // populated serving store
+      await jest.advanceTimersByTimeAsync(fetchResolveDelayMs);
+      expect(fetchCallCount).toBe(2);
+      expect(fetchResolveCount).toBe(2);
+      expect(client.getStringAssignment(flagKey1, 'subject', {}, 'default-value')).toBe(
+        updateOnFetch === 'empty' ? 'control' : 'default-value',
```
leoromanovsky, line 780: 💯 💯 💯

src/index.spec.ts
```diff
@@ -528,6 +523,285 @@ describe('initialization options', () => {
...
+    expect(fetchResolveCount).toBe(1);
+    expect(client.getStringAssignment(flagKey, 'subject', {}, 'default-value')).toBe('control');
+
+    // Should not be tripped up by failed fetch
+    await jest.advanceTimersByTimeAsync(fetchResolveDelayMs);
+    expect(fetchResolveCount).toBe(2);
+    expect(client.getStringAssignment(flagKey, 'subject', {}, 'default-value')).toBe('control');
+  });
+
+  it('Ignores cache if fetch finishes first', async () => {
```
leoromanovsky, line 651: the android test 😄

leoromanovsky (approved):
`index.spec.ts` is a work of art - insane how you got coverage over all cases.

are you considering dogfooding on eppo with a RC build?

src/isolated-hybrid.store.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export type ServingStoreUpdateStrategy = 'always' | 'expired' | 'empty';
+
+/**
+ * Extension of the HybridConfigurationStore that allows isolating the serving store from updates.
+ * This could be useful when you want to avoid updating the serving store to maintain greater
+ * consistency within a session, but still want to update the cache for a quick, updated
+ * initialization the next session.
+ */
+export class IsolatedHybridConfigurationStore<T> extends HybridConfigurationStore<T> {
```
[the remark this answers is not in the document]
aarsilv: Good point--will do!

src/isolated-hybrid.store.ts
```diff
@@ -0,0 +1,44 @@
+import { HybridConfigurationStore, IAsyncStore, ISyncStore } from '@eppo/js-client-sdk-common';
+
+export type ServingStoreUpdateStrategy = 'always' | 'expired' | 'empty';
```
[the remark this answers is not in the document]
aarsilv: Eh, there is no real reuse of that default besides the default param for the constructor. After that the `always` constant is only really used in the context of tests checking that specific strategy (vs. a default strategy)

src/isoldated-hybrid.store.spec.ts
aarsilv: good spot!

PR #81, opened 2024-06-19
feat: [add syncInit method] (FF-2431)

leoromanovsky:
Fixes: #__issue__

## Motivation and Context

Users need to be able to bootstrap the JS SDK from configuration obtained from their server, or another external process. The format of this configuration might or might not be obfuscated.

Furthermore users need to be able to instantiate the SDK synchronously. 

## Description

* adds a new public function `syncInit` - it accepts a map of flags in both obfuscated and not-obfuscated formats. pairs with the `setIsObfuscated` parameter to make the SDK understand the incoming data.
* Augments the README with instructions for how to use this method.

## How has this been tested?

New unit tests to validate behavior of providing obfuscated and not-obfuscated configuration.

[//]: # (OPTIONAL)


docs/js-client-sdk.eppojsclient.getintegerassignment.md
```diff
@@ -16,7 +16,7 @@ getIntegerAssignment(flagKey: string, subjectKey: string, subjectAttributes: Rec
 |  --- | --- | --- |
 |  flagKey | string |  |
 |  subjectKey | string |  |
-|  subjectAttributes | Record&lt;string, any&gt; |  |
+|  subjectAttributes | Record&lt;string, AttributeType&gt; |  |
```
leoromanovsky, line 19: I don't get why this is changing - I'm assuming someone else is checking in doc changes without this?

src/index.spec.ts
```diff
@@ -299,14 +364,42 @@ describe('EppoJSClient E2E test', () => {
...
+        [flagKey]: mockNotObfuscatedFlagConfig,
+      },
+    });
+
+    expect(client.getStringAssignment(flagKey, 'subject-10', {}, 'default-value')).toEqual(
+      'variant-1',
+    );
+  });
+});
+
```
leoromanovsky, line 394: new tests for the functionality being added in this PR

src/index.spec.ts
```diff
@@ -47,7 +55,64 @@ const obfuscatedFlagKey = md5Hash(flagKey);
 const allocationKey = 'traffic-split';
 const obfuscatedAllocationKey = base64Encode(allocationKey);
 
-const mockUfcFlagConfig: Flag = {
```
leoromanovsky, line 50: Renamed this to be called `obfuscated` for clarity

src/index.ts
```diff
@@ -110,6 +110,16 @@ export interface IClientConfig {
...
+export interface IClientConfigSync {
+  flagsConfiguration: Record<string, Flag>;
+
+  assignmentLogger?: IAssignmentLogger;
+
+  isObfuscated?: boolean;
+
+  throwOnFailedInitialization?: boolean;
+}
+
```
leoromanovsky: These are the possible configuration options for calling the `initSync` method - note that no SDK key is provided as it will be operating in off-line mode.

Assuming that on the server the user stringified the configurations to a JSON string, they will need to parse it back into the object using the types we export in the SDK.

src/index.ts
```diff
@@ -209,6 +219,60 @@ export function buildStorageKeySuffix(apiKey: string): string {
...
+    // since the purpose of this method is to bootstrap the SDK from an external source,
+    // which is likely a server that has not-obfuscated flag values.
+    EppoJSClient.instance.setIsObfuscated(config.isObfuscated ?? false);
+
+    if (config.assignmentLogger) {
+      EppoJSClient.instance.setLogger(config.assignmentLogger);
+    }
+
+    // There is no SDK key in the offline context.
+    const storageKeySuffix = buildStorageKeySuffix('no_api_key');
```
leoromanovsky, line 255: Got another idea for handling this?

src/index.ts
```diff
@@ -209,6 +219,60 @@ export function buildStorageKeySuffix(apiKey: string): string {
...
+
+    // As this is a synchronous initialization,
+    // we are unable to call the async `init` method on the assignment cache
+    // which loads the assignment cache from the browser's storage.
+    // Therefore there is no purpose trying to use a persistent assignment cache.
+    const assignmentCache = assignmentCacheFactory({
+      storageKeySuffix,
+      forceMemoryOnly: true,
+    });
+    EppoJSClient.instance.useCustomAssignmentCache(assignmentCache);
```
leoromanovsky, line 265: There is still value to have an in-memory cache as it will remove duplicates for the duration of the session. Perhaps we can think creatively how to support this.

src/index.ts
```diff
@@ -225,6 +289,8 @@ export async function init(config: IClientConfig): Promise<IEppoClient> {
     instance.stopPolling();
     // Set up assignment logger and cache
     instance.setLogger(config.assignmentLogger);
+    // Default to obfuscated mode when requesting configuration from the server.
+    instance.setIsObfuscated(true);
```
leoromanovsky, line 296: Making it clear here.

README.md
```diff
@@ -85,7 +84,26 @@ The `init` function accepts the following optional configuration arguments.
...
 
+When this function returns the SDK is ready for assignments.
+
+```javascript
+import { initSync, Flag, ObfuscatedFlag } from "@eppo/js-client-sdk";
+
+// configuration from your server SDK
+const configurationJsonString: string = getConfigurationFromServer();
+// The configuration will be not-obfuscated from your server SDK. If you have obfuscated flag values, you can use the `ObfuscatedFlag` type.
+const flagsConfiguration: Record<string, Flag | ObfuscatedFlag> = JSON.parse(configurationJsonString);
```
felipecsl, line 99: nit: I think i would omit the types for brevity since they are inferred anyways

docs/js-client-sdk.chromestorageengine._constructor_.md
```diff
@@ -0,0 +1,21 @@
...
+[Home](./index.md) &gt; [@eppo/js-client-sdk](./js-client-sdk.md) &gt; [ChromeStorageEngine](./js-client-sdk.chromestorageengine.md) &gt; [(constructor)](./js-client-sdk.chromestorageengine._constructor_.md)
+
+## ChromeStorageEngine.(constructor)
+
+Constructs a new instance of the `ChromeStorageEngine` class
+
+**Signature:**
+
+```typescript
+constructor(storageMap: ChromeStorageAsyncMap<string>, storageKeySuffix: string);
```
felipecsl, line 12: these should have been generated from a previous commit. is there a command that I should have run?

src/index.ts
```diff
@@ -209,6 +222,60 @@ export function buildStorageKeySuffix(apiKey: string): string {
...
+ * The purpose is for use-cases where the configuration is available from an external process
+ * that can bootstrap the SDK.
+ *
+ * This method should be called once on application startup.
+ *
+ * @param config - client configuration
+ * @returns a singleton client instance
+ * @public
+ */
+export function initSync(config: IClientConfigSync): IEppoClient {
```
felipecsl, line 237: `syncInit` instead?

src/cache/assignment-cache-factory.ts
```diff
@@ -8,14 +8,21 @@ import { LocalStorageAssignmentCache } from './local-storage-assignment-cache';
...
   chromeStorage,
   storageKeySuffix,
 }: {
+  forceMemoryOnly?: boolean;
   storageKeySuffix: string;
   chromeStorage?: chrome.storage.StorageArea;
 }): AssignmentCache {
-  const hasLocalStorage = hasWindowLocalStorage();
   const simpleCache = new SimpleAssignmentCache();
+  if (forceMemoryOnly) {
```
felipecsl, line 20: instead of doing this, which makes this method more complex (unecessarily, I think), i'd just instantiate `SimpleAssignmentCache` directly from `syncInit`

src/index.ts
felipecsl: just use a random string instead, I think that would be cleaner

felipecsl (commented):
Main feedback is around the assignment cache/config cache usage, otherwise looks good to me

docs/js-client-sdk.chromestorageengine._constructor_.md
sameerank: yes you can run `yarn run docs`, but it's also created by the pre-commit script

https://github.com/Eppo-exp/js-client-sdk/blob/e2f01d438686423609b726adfc49525f68f46d47/package.json#L17

Although there's a mis-ordering of events so the commit ends up happening without including the generated docs, which is probably why they were left out. I'm increasingly inclined to removing these autogenerated docs altogether though. I don't believe anyone views them

felipecsl: is it this stuff? https://eppo-exp.github.io/js-client-sdk/js-client-sdk.html

src/index.ts
leoromanovsky: Honestly I don't like either - seems a little on the nose: should be obvious its sync from the `lack` of `async` in the function definition. Let me ponder something more elegant...

felipecsl: yeah I think we may need a completely different name here to capture the fact that this initializes "offline" and doenst make any network requests. Maybe `initFromPrefetchedConfig`? Something more explicit would be better IMHO

sameerank: `offlineInit`?

src/index.ts
```diff
@@ -209,6 +222,60 @@ export function buildStorageKeySuffix(apiKey: string): string {
...
+    const memoryOnlyConfigurationStore = configurationStorageFactory({
+      forceMemoryOnly: true,
+    });
+    memoryOnlyConfigurationStore.setEntries(config.flagsConfiguration);
+    EppoJSClient.instance.setConfigurationStore(memoryOnlyConfigurationStore);
+
+    // Allow the caller to override the default obfuscated mode, which is false
+    // since the purpose of this method is to bootstrap the SDK from an external source,
+    // which is likely a server that has not-obfuscated flag values.
+    EppoJSClient.instance.setIsObfuscated(config.isObfuscated ?? false);
```
sameerank, line 249: Just a suggestion. Setting the defaults at the top of the function might help with readability

```
const isObfuscated = config.isObfuscated ?? false;
const throwOnFailedInitialization = config.throwOnFailedInitialization ?? true;
```

sameerank (approved):

src/index.ts
leoromanovsky: The only thing I don't love about this:

* generate a random string in JS seems like so much code `[...Array(8)].map(() => Math.random().toString(36)[2]).join('');` but that's fine
* if we move to having a persistent store i don't want to generate a different random string each initialization which would make the store useless. perhaps a different static string like `offline` is more clear?

felipecsl: yeah I think either should be fine, as long as the intention is clear. could be as simple as extracting to a function like `generateStorageCacheKeySuffixForOffline` or something like that

src/index.ts
leoromanovsky: That's a good one, thanks!

src/cache/assignment-cache-factory.ts
leoromanovsky: Yea - I don't have strong feelings about it either way. On the one hand this method becomes the central way to instantiate caches without the caller needing to know anything about the `SimpleAssignmentCache` besides saying they want an `inMemory`.

felipecsl: i feel like it's an unnecessary abstraction that adds complexity. not gonna block on it, but it's my personal preference ;)

PR #86, opened 2024-06-21
remove doc generation

leoromanovsky:
---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context

These generated docs seem pretty low value, they just explain the types which should be self-evident, and are just causing our PRs to be larger in size then they need to be.

## Description

* remove `docs` directory
* remove scripts and dependencies that generate them

## How has this been tested?

[//]: # (OPTIONAL)


sameerank (approved):
🪓

felipecsl:
While I understand these were not adding a lot of value as is, I do think some sort of SDK API reference is important to have and we should revisit that in the future maybe. I don't think that just having the official Eppo docs is enough. As an example, here are Stripe's JS SDK API reference: https://docs.stripe.com/terminal/references/api/js-sdk#stripeterminal-create

PR #95, opened 2024-08-01
feat: [do not re-init and perform a network request if the singleton …

leoromanovsky:
…has already been init] (FF-2921)

---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context

If a user calls `init` repeatedly, the client will issue multiple network requests. This is not desirable behavior as the configuration and assignment cache are also reset.

To support users explicitly reloading configuration we should offer a dedicated refetch method.

## Description

* Default behavior change: if the singleton already has been initialized, defined by the `init` method completing (not taking into account the `AsyncStore` or others), return it immediately without re-instantiating.
* Offers a new flag called `forceReinitialize` - setting to `true` re-instantiates the client. I'm not convinced this is needed for users but is useful for tests.
* Bumped to `3.5.0` since this is new functionality - you can argue that functionality is "breaking" but I view this as a correction. We could even argue to publish it as a bug fix.

## How has this been tested?

* New unit tests.

## Question for reviewers

1. Should we publish as a patch release?

[//]: # (OPTIONAL)

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -191,6 +191,7 @@ describe('EppoJSClient E2E test', () => {
       apiKey,
       baseUrl,
       assignmentLogger: mockLogger,
+      forceReinitialize: true,
     });
   });
 
@@ -335,6 +336,7 @@ describe('EppoJSClient E2E test', () => {
         apiKey,
         baseUrl,
         assignmentLogger: mockLogger,
+        forceReinitialize: true,
       });
     });
 
@@ -778,6 +780,7 @@ describe('initialization options', () => {
       apiKey,
       baseUrl,
       assignmentLogger: mockLogger,
+      forceReinitialize: true,
     });
 
     expect(fetchCallCount).toBe(1);
@@ -790,6 +793,7 @@ describe('initialization options', () => {
       baseUrl: 'https://thisisabaddomainforthistest.com',
       assignmentLogger: mockLogger,
       useExpiredCache: true,
+      forceReinitialize: true,
     });
 
     // Should serve assignment from cache before fetch even fails
@@ -809,6 +813,7 @@ describe('initialization options', () => {
         baseUrl: 'https://thisisabaddomainforthistest.com',
         assignmentLogger: mockLogger,
         useExpiredCache: true,
+        forceReinitialize: true,
       }),
     ).rejects.toThrow();
   });
@@ -909,6 +914,7 @@ describe('initialization options', () => {
         baseUrl,
         assignmentLogger: mockLogger,
         updateOnFetch,
+        forceReinitialize: true,
       });
 
       expect(fetchCallCount).toBe(1);
@@ -926,6 +932,7 @@ describe('initialization options', () => {
         assignmentLogger: mockLogger,
         updateOnFetch,
         useExpiredCache: true,
+        forceReinitialize: true,
       });
 
       // Should serve assignment from cache before fetch completes
@@ -955,6 +962,7 @@ describe('initialization options', () => {
         assignmentLogger: mockLogger,
         updateOnFetch,
         maxCacheAgeSeconds: fetchResolveDelayMs * 10,
+        forceReinitialize: true,
       });
 
       // No fetch will have been kicked off because of valid cache; previously fetched values will be served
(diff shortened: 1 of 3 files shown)


src/index.spec.ts
```diff
@@ -191,6 +191,7 @@ describe('EppoJSClient E2E test', () => {
       apiKey,
       baseUrl,
       assignmentLogger: mockLogger,
+      forceReinitialize: true,
```
leoromanovsky, line 194: A lot of this is pretty messy - ideally we would not need to do this in tests but the singletons persist across test cases. Maybe this trick is what's needed for an isolated test:

```
      jest.isolateModules(() => {
        // Isolate and re-require so that the static instance is reset to it's default state
        // eslint-disable-next-line @typescript-eslint/no-var-requires
        const reloadedModule = require('./index');
        init = reloadedModule.init;
        getInstance = reloadedModule.getInstance;
      });
```

src/index.spec.ts
```diff
@@ -871,6 +932,7 @@ describe('initialization options', () => {
         assignmentLogger: mockLogger,
         updateOnFetch,
         useExpiredCache: true,
+        forceReinitialize: true,
```
leoromanovsky, line 935: some of the test cases here could be incorrect. this one:

```
// Init again with an expired cache so a fetch will kicked off too
// We allow the expired cache, so it will succeed and fetch will be delayed
```

might be contrary to the singleton approach - or perhaps we need to check more conditions than just `initialized`

src/index.ts
```diff
@@ -381,8 +386,23 @@ export async function init(config: IClientConfig): Promise<EppoClient> {
...
   let initializationError: Error | undefined;
   const instance = EppoJSClient.instance;
-  const { apiKey, persistentStore, baseUrl, maxCacheAgeSeconds, updateOnFetch } = config;
+  const { apiKey, persistentStore, baseUrl, maxCacheAgeSeconds, updateOnFetch, forceReinitialize } =
+    config;
   try {
+    if (EppoJSClient.initialized) {
+      if (forceReinitialize) {
+        applicationLogger.warn(
+          'Eppo SDK is already initialized. Since forceReinitialize is true, reinitializing.',
```
felipecsl, line 395: i'd change this to an `info` instead of a `warn`. also minor rewording: `Eppo SDK is already initialized. Reinitializing since forceReinitialize is true.`

src/index.ts
```diff
@@ -381,8 +386,23 @@ export async function init(config: IClientConfig): Promise<EppoClient> {
...
+        );
+        EppoJSClient.initialized = false;
+      } else {
+        applicationLogger.warn(
+          'Eppo SDK is already initialized. Since forceReinitialize is false, returning existing instance.',
```
felipecsl, line 400: `Eppo SDK is already initialized, skipping reinitialization since forceReinitialize is false`

felipecsl (approved):

src/index.spec.ts
aarsilv: I think it's nice to be able to force re-initialize, for tests and possibly other applications

src/index.spec.ts
```diff
@@ -494,6 +496,61 @@ describe('initialization options', () => {
...
+    });
+
+    await init({
+      apiKey,
+      baseUrl,
+      assignmentLogger: mockLogger,
+      forceReinitialize: true,
+    });
+
+    expect(callCount).toBe(2);
```
aarsilv, line 551: 👍

src/index.spec.ts
aarsilv: I think we were using multiple initializations as a hack to check populating and then using the cache. If we want, we could rework tests to check cache contents and set up an already-populated cache another way but I think this is probably fine for now.

src/index.ts
aarsilv: ☝️ I like this reword

aarsilv (approved):
Thanks for doing this!

src/index.ts
leoromanovsky: TBH none of these logs are getting emitted anyways since this is the browser 😄

felipecsl: hmm would be nice to use a tool like `debug` so consumers of the library can choose whether or not to emit console logging

src/index.spec.ts
felipecsl: ah nvm it looks like we use pino exactly for that

PR #96, opened 2024-08-21
quality: parameterize branch names and make test

typotter:
🎟️ Fixes FF-3092 towards FF-3085

👯‍♂️ **Related PRs**
- [sdk-test-data#58](https://github.com/Eppo-exp/sdk-test-data/pull/58)
- [java-server-sdk#75](https://github.com/Eppo-exp/java-server-sdk/pull/75)
- [python-sdk#65](https://github.com/Eppo-exp/python-sdk/pull/65)

### Motivation
Changes are often made to the [sdk-test-data](https://github.com/Eppo-exp/sdk-test-data) repository to capture new behaviours, bugs and edge cases. When these changes are pushed to `main`, the SDKs are cloned locally (locally to the github action running) and their respective tests are run. These tests are set up and run by copies of the SDK test workflows - see [sdk-test-data workflow](https://github.com/Eppo-exp/sdk-test-data/blob/main/.github/workflows/test-sdks.yml). There are a number of limitations to this setup:

- Test steps are copied from the SDK’s respective workflows; changes to the SDK test workflows need to be replicated in sdk-test-data and this is not obvious to devs
- The SDK tests are not able to run against in-flight changes to `sdk-test-data`
- When new test-data is committed that breaks an SDK, that breakage is not surfaced in the SDK’s repository until some action triggers its testing workflow (on demand, on pull-request, etc.).

### Description of Changes
_This change_
🚀 - Each SDK's testing workflow is enhanced into a [reusable workflow](https://docs.github.com/en/actions/sharing-automations/reusing-workflows), exposing parameters for SDK branch and the sdk-test-data branch to use in testing.
- The test workflow runs using the main branch of sdk-test-data on all main pushes and Pull Requests using the PR's branch

_External to this Change_
[sdk-test-data](https://github.com/Eppo-exp/sdk-test-data/blob/main/.github/workflows) get two testing workflows.
1. ♻️  "Local Testing"- For all pull request changes, the "Local Testing" workflow calls the reusable SDK workflows, test results are recorded only in the sdk-test-data action. This is run using the main SDK branch and the "current" branch of workflow, i.e. the pull request branch. (SDK repo does not see/is not notified of test failures during PR lifecycle, only on push to main)
2. ♻️ 🧑‍💻 "Remote Testing" - On all pushes to the main branch of sdk-test-data, the SDK testing workflows are triggered to run within their respective repositories, alerting all subscribers of failed runs (repo is red/green).

### DIFF
--- a/.github/workflows/lint-test-sdk.yml
+++ b/.github/workflows/lint-test-sdk.yml
@@ -4,11 +4,37 @@ on:
     paths:
       - '**/*'
 
+  workflow_dispatch:
+
+  workflow_call:
+    inputs:
+      test_data_branch:
+        type: string
+        description: The branch in sdk-test-data to target for testcase files
+        required: false
+        default: main
+      sdk_branch:
+        type: string
+        description: The branch of the SDK to test
+        required: false
+
+env:
+  SDK_BRANCH_NAME: ${{ inputs.sdk_branch  || github.head_ref || github.ref_name || 'main' }}
+  TEST_DATA_BRANCH_NAME: ${{ inputs.test_data_branch || 'main' }}
+        
 jobs:
   lint-test-sdk:
     runs-on: ubuntu-latest
     steps:
+      - name: Display Testing Details
+        run: |
+          echo "Running SDK Test using"
+          echo "Test Data: sdk-test-data@${TEST_DATA_BRANCH_NAME}"
+          echo "SDK Branch: js-client-sdk@${SDK_BRANCH_NAME}"
       - uses: actions/checkout@v2
+        with:
+          repository: Eppo-exp/js-client-sdk
+          ref: ${{ env.SDK_BRANCH_NAME }}
       - name: Use Node.js
         uses: actions/setup-node@v1
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/lint-test-sdk.yml
```diff
@@ -4,11 +4,37 @@ on:
...
+        description: The branch in sdk-test-data to target for testcase files
+        required: false
+        default: main
+      sdk_branch:
+        type: string
+        description: The branch of the SDK to test
+        required: false
+
+env:
+  SDK_BRANCH_NAME: ${{ inputs.sdk_branch  || github.head_ref || github.ref_name || 'main' }}
```
greghuels, line 22: Do we want a fallback to `main` here? I would think merges to main would be covered by `github.ref_name`, wouldn't it? If that's the case, it might make sense to throw an error if `inputs.sdk_branch  || github.head_ref || github.ref_name` result in an empty `$SDK_BRANCH_NAME` since that would indicate something was wrong with the configuration.

greghuels (approved):
Looking great so far! Nothing blocking - just a question about falling back to 'main'

.github/workflows/lint-test-sdk.yml
typotter: You are right that the `main` fallback here is superfluous. Looking at _Actions/checkout_ readme and code, the ref will be computed as the triggering branch or the default branch if `ref` is omitted or is a blank string.

The checkout action will throw an error when the refspec doesn't exist, stopping the workflow and displays an error message indicating as much. I'm not sure throwing our own error here would add much, to be honest, but I am trying to get these all hooked up so I can move on to the SDK packaging and integration testing so I may be biased.

greghuels: I don't mean we need to create our own error. I just meant that it might give a false sense of security if we fallback to main when it might indicate a larger problem with our configuration. In any case, it's not a blocker; just something I wanted to point out as an improvement opportunity in a subsequent iteration.

greghuels: > The checkout action will throw an error when the refspec doesn't exist

Ah, right. This is a good callout. Okay, cool.

PR #104, opened 2024-11-27
fix: init should not require event ingestion url in sdk key

sameerank:
---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context

I was running a script to try out the `init` function and saw the following exception thrown when I invoked `init` with a pre-existing SDK key.

```
Error during initialization test: Error: Unable to parse Event ingestion URL from SDK key
    at newDefaultEventDispatcher (/.../js-client-sdk/node_modules/@eppo/js-client-sdk-common/src/events/default-event-dispatcher.ts:138:11)
```

I think the SDK should continue to support old keys, or at least handle the exception so it doesn't get thrown outside the SDK's scope

## Description

Here, I propose wrapping the event dispatcher instantiation with a try-catch. Other approaches, like handling this in `js-client-sdk-common` may be valid too

## How has this been tested?

[//]: # (OPTIONAL)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -388,7 +388,15 @@ export async function init(config: IClientConfig): Promise<EppoClient> {
       skipInitialPoll: skipInitialRequest,
     };
     instance.setConfigurationRequestParameters(requestConfiguration);
-    instance.setEventDispatcher(newEventDispatcher(apiKey));
+
+    try {
+      instance.setEventDispatcher(newEventDispatcher(apiKey));
+    } catch (eventDispatcherError) {
+      applicationLogger.warn(
+        'Eppo SDK encountered an error initializing the event dispatcher, continuing initialization',
+        eventDispatcherError,
+      );
+    }
 
     // We have two at-bats for initialization: from the configuration store and from fetching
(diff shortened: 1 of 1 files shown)


aarsilv (approved):
Big fan of SDK failure modes, thanks for thinking through this!

src/index.ts
```diff
@@ -388,7 +388,16 @@ export async function init(config: IClientConfig): Promise<EppoClient> {
       skipInitialPoll: skipInitialRequest,
     };
     instance.setConfigurationRequestParameters(requestConfiguration);
-    instance.setEventDispatcher(newEventDispatcher(apiKey));
+
+    // e.g. newDefaultEventDispatcher throws an error if it cannot parse a valid event ingestion URL from the SDK key
+    try {
+      instance.setEventDispatcher(newEventDispatcher(apiKey));
```
felipecsl, line 394: fixed in https://github.com/Eppo-exp/js-sdk-common/pull/143 to no longer throw
I'd vote removing this try/catch since it should never throw after that change

sameerank: Sounds good. I'll close this out

PR #115, opened 2024-12-04
fix: Set `process.env.LOG_LEVEL` to `null` in webpack config + Buffer polyfill

felipecsl:
---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context
Fix `process is not defined` error

## Description
https://github.com/Eppo-exp/js-sdk-common/pull/136 introduced a change to `applicationLogger` that added a check for `process.env.LOG_LEVEL`.
This doesn't work on the browser because `process` only exists in the Node environment.
Although there's a tricky reason why the original `(process.env.NODE_ENV === 'production' ? 'warn' : 'info')` expression didn't already fail, which is that webpack resolves the entire expression to simply `warn` since it resolves `process.env.NODE_ENV` to `production` at build time.
I've worked around it by doing a similar preprocessing step for `LOG_LEVEL` by adding a `webpack.DefinePlugin` to the webpack config that sets it to null.
This causes webpack to simplify the entire line to just `level: "warn"`.

## How has this been tested?
`yarn webpack`
Inspect the generated file `dist/eppo-sdk.js`. It should have a line line like below:
```
t.loggerPrefix = "[Eppo SDK]", t.logger = (0, n.default)({level: "warn", browser: {disabled: !0}})
```
Create a `index.html` file with the following contents:
```
<script src="file:///absolute/path/to/js-client-sdk/dist/eppo-sdk.js"></script>
<script>
    console.log(window.eppo);
</script>
```
Load this page in the browser and there should be no errors in the dev tools console.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "webpack-cli": "^4.10.0"
   },
   "dependencies": {
-    "@eppo/js-client-sdk-common": "^4.5.3"
+    "@eppo/js-client-sdk-common": "^4.5.4"
   },
   "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@eppo/js-client-sdk",
-  "version": "3.8.1",
+  "version": "3.8.2",
   "description": "Eppo SDK for client-side JavaScript applications",
   "main": "dist/index.js",
   "files": [
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -2,6 +2,7 @@
 const path = require('path');
 
 const TerserPlugin = require('terser-webpack-plugin');
+const webpack = require('webpack');
 
 module.exports = {
   entry: './src/index.ts',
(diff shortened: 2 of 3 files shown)


leoromanovsky (approved):
thanks for testing thoroughly

sameerank (approved):
Great work investigating! Do you want to patch the version in this PR too to publish the fix?

greghuels (changes requested):
Screenshot 2024-12-04 at 4 58 56 AM

I'm still getting Buffer is not defined, so I don't think this is the full solution. It looks like it's coming from https://github.com/Eppo-exp/js-sdk-common/blob/0587b36dc4547a29fbfd351ad3a46954d3263684/src/events/sdk-key-decoder.ts#L10. I think we need to update that file with the `decodeBase64` function in `src/obfuscation.ts`.

felipecsl:
> Screenshot 2024-12-04 at 4 58 56 AM
> I'm still getting Buffer is not defined, so I don't think this is the full solution. It looks like it's coming from https://github.com/Eppo-exp/js-sdk-common/blob/0587b36dc4547a29fbfd351ad3a46954d3263684/src/events/sdk-key-decoder.ts#L10. I think we need to update that file with the `decodeBase64` function in `src/obfuscation.ts`.

We already added a polyfill for buffer, so you should not be running into that. how are you testing it? did you run `yarn install`?

felipecsl:
> Great work investigating! Do you want to patch the version in this PR too to publish the fix?

Yup i'll bump the version and release

greghuels:
> how are you testing it? did you run `yarn install`?

I'm using the `scripts/local-install.sh` in this repo to install the js-client-sdk into a basic Next.js app I generated locally. I did yarn install, yes.

felipecsl:
Screenshot 2024-12-04 at 09 18 20

felipecsl:
Bumped shared sdk, retested everything, good to go!

greghuels (approved):
it works now!

PR #117, opened 2024-12-05
feat: create offlinePrecomputedInit

sameerank:
---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context

## Description

Adds a method for `offlinePrecomputedInit`

Depends on https://github.com/Eppo-exp/js-sdk-common/pull/163 because we need a way to persist the subject key and attributes without relying on request parameters, since the offline init doesn't make requests

## How has this been tested?

[//]: # (OPTIONAL)


package.json
```diff
@@ -59,7 +59,7 @@
     "webpack-cli": "^4.10.0"
   },
   "dependencies": {
-    "@eppo/js-client-sdk-common": "^4.5.4"
+    "@eppo/js-client-sdk-common": "^4.6.1"
```
sameerank, line 62: Depends on this to be published https://github.com/Eppo-exp/js-sdk-common/pull/163

src/i-client-config.ts
```diff
@@ -78,6 +78,14 @@ export interface IPrecomputedClientConfig extends IBaseRequestConfig {
   subjectAttributes?: Record<string, AttributeType>;
 }
 
+/**
+ * Configuration for Eppo precomputed client initialization
+ * @public
+ */
+export interface IPrecomputedClientConfig extends IBaseRequestConfig {
```
sameerank, line 85: Breaking change to the interface of `precomputedInit` but assuming it is safe because we are currently the only users of this function

sameerank:
Marking as ready for review because I expect the tests to pass once the js common dependency is published

src/index.ts
```diff
@@ -624,3 +621,70 @@ function newEventDispatcher(sdkKey: string): EventDispatcher {
...
+ *
+ * The purpose is for use-cases where the precomputed assignments are available from an external process
+ * that can bootstrap the SDK.
+ *
+ * This method should be called once on application startup.
+ *
+ * @param config - client configuration
+ * @returns a singleton precomputed client instance
+ * @public
+ */
```
aarsilv, line 636: Comment needing updating to speak to the interface?

aarsilv (approved):
Pieces coming together 💪

src/index.ts
sameerank: Great catch! I see I didn't edit it enough

src/index.ts
```diff
@@ -17,7 +17,10 @@ import {
   ObfuscatedFlag,
   BoundedEventQueue,
   validation,
+  PrecomputedFlag,
+  NamedEventQueue,
 } from '@eppo/js-client-sdk-common';
+import { Event } from '@eppo/js-client-sdk-common/src/events/event-dispatcher';
```
sameerank, line 23: @felipecsl I needed to do an `as NamedEventQueue<Event>` in order get an event type to work with `newDefaultEventDispatcher`. Am I doing this right? Are you okay with including `Event` in the public exports?

felipecsl: ah yeah we should be exporting that! i'll send a PR!

src/index.ts
```diff
@@ -622,5 +621,76 @@ function newEventDispatcher(sdkKey: string): EventDispatcher {
     { isOffline: () => false, onNetworkStatusChange: () => {} };
   const networkStatusListener =
     typeof window !== 'undefined' ? new BrowserNetworkStatusListener() : emptyNetworkStatusListener;
-  return newDefaultEventDispatcher(eventQueue, networkStatusListener, sdkKey);
+  return newDefaultEventDispatcher(
+    eventQueue as NamedEventQueue<Event>,
```
felipecsl, line 625: You could also explicitly type it on line 616 above, that's slightly cleaner IMHO

src/index.ts
felipecsl: https://github.com/Eppo-exp/js-sdk-common/pull/171

leoromanovsky (commented):
I haven't reviewed this yet but since it has a ✅ please remember to test this build by compiled the production package.

Makefile
```diff
@@ -27,14 +27,15 @@ help: Makefile
 testDataDir := test/data/
 tempDir := ${testDataDir}temp/
 gitDataDir := ${tempDir}sdk-test-data/
-branchName := main
+branchName := sameeran/ff-3687-obfuscated-precomputed-json
```
sameerank, line 30: Temporarily using this branch for testing until https://github.com/Eppo-exp/sdk-test-data/pull/89 gets an approval 🙏

src/index.spec.ts
```diff
@@ -1179,6 +1139,50 @@ describe('EppoPrecomputedJSClient E2E test', () => {
   });
 });
 
+describe('offlinePrecomputedInit', () => {
+  let mockLogger: IAssignmentLogger;
+  let precomputedConfigurationWire: string;
+
+  beforeAll(() => {
+    precomputedConfigurationWire = readMockPrecomputedResponse(MOCK_PRECOMPUTED_RESPONSE_FILE);
```
sameerank, line 1147: Currently, tests use the unobfuscated, unstringified `ConfigurationWire` format https://github.com/Eppo-exp/sdk-test-data/blob/8ee55117b7c7f5d02b2997629af0f0fa93e4e97c/configuration-wire/precomputed-v1.json

In a follow up PR I will add support for parsing json strings and handling obfuscated fields

src/index.ts
```diff
@@ -605,6 +603,92 @@ export async function precomputedInit(
...
+    fetchedAt: string;
+    response: {
+      createdAt: string;
+      format: FormatEnum;
+      obfuscated: boolean;
+      environment: Environment;
+      flags: Record<string, PrecomputedFlag>;
+    };
+  };
+}
```
sameerank, line 638: Will replace with an exported interface from https://github.com/Eppo-exp/js-sdk-common/pull/160

src/index.ts
```diff
@@ -605,6 +603,93 @@ export async function precomputedInit(
...
+  } catch (error) {
+    applicationLogger.warn(
+      'Eppo SDK encountered an error initializing precomputed client, assignment calls will return the default value and not be logged',
+    );
+    if (throwOnFailedInitialization) {
+      throw error;
+    }
+  }
+
+  EppoPrecomputedJSClient.instance.setIsObfuscated(parsedResponse.obfuscated);
```
typotter, line 688: `parsedResponse` also has the salt needed to hash flagKeys in `getFlag`

src/index.spec.ts
```diff
@@ -1205,7 +1214,7 @@ describe('EppoClient config', () => {
     const retryManager = eventDispatcher['retryManager'];
     const batchProcessor = eventDispatcher['batchProcessor'];
     expect(eventDispatcher['deliveryIntervalMs']).toEqual(1);
-    expect(batchProcessor['batchSize']).toEqual(5);
+    expect(batchProcessor['batchSize']).toEqual(100);
```
sameerank, line 1217: cc @felipecsl I had to change this because 5 is below the minimum allowed batch size

https://github.com/Eppo-exp/js-sdk-common/blob/b1048a32587a809bed351af07185592c9f774150/src/events/batch-event-processor.ts#L4-L13

src/index.ts
sameerank: Yup using the salt now

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@eppo/js-client-sdk",
-  "version": "3.8.3",
+  "version": "3.9.0-alpha.0",
```
sameerank, line 3: ⚠️ Do not publish this version for dogfooding until 

1. The edge compute endpoint is published with updated hashing
2. The corresponding md5 hasher in js-client-sdk-common is updated https://github.com/Eppo-exp/js-sdk-common/pull/164/files#r1887291803 (`.appendBinary` --> `.append`)
3. And 4.7.0-alpha.1 version of js-client-sdk-common has been released and the dependency in this file has been updated to match

src/index.ts
```diff
@@ -605,6 +607,89 @@ export async function precomputedInit(
...
+      .setEntries(parsedResponse.flags)
+      .catch((err) =>
+        applicationLogger.warn('Error setting precomputed assignments for memory-only store', err),
+      );
+
+    const subjectAttributes: Attributes = convertContextAttributesToSubjectAttributes(
+      contextAttributes ?? { numericAttributes: {}, categoricalAttributes: {} },
+    );
+
+    EppoPrecomputedJSClient.instance.setSubjectSaltAndPrecomputedFlagStore(
```
sameerank, line 669: Based on these PR comments 

https://github.com/Eppo-exp/js-sdk-common/pull/164#discussion_r1905896857
https://github.com/Eppo-exp/js-sdk-common/pull/164#discussion_r1905899272

I'll see if I can make this method private/not use it here and have these set through the constructor

src/index.ts
```diff
@@ -597,9 +589,95 @@ export async function precomputedInit(
...
+ *
+ * The purpose is for use-cases where the precomputed assignments are available from an external process
+ * that can bootstrap the SDK.
+ *
+ * This method should be called once on application startup.
+ *
+ * @param config - client configuration
+ * @returns a singleton precomputed client instance
+ * @public
+ */
```
rasendubi, line 619: The documentation comment does not belong to interface below.

src/index.ts
```diff
@@ -597,9 +589,95 @@ export async function precomputedInit(
...
+export function offlinePrecomputedInit(
+  config: IPrecomputedClientConfigSync,
+): EppoPrecomputedClient {
+  const throwOnFailedInitialization = config.throwOnFailedInitialization ?? true;
+
+  const configurationWire: IConfigurationWire = JSON.parse(config.precomputedConfigurationWire);
+  if (!configurationWire.precomputed) {
+    applicationLogger.error('Invalid precomputed configuration wire');
+    return EppoPrecomputedJSClient.instance;
+  }
```
rasendubi, line 647: This does not comply to `throwOnFailedInitialization`

src/index.ts
```diff
@@ -597,9 +589,95 @@ export async function precomputedInit(
...
+  const { subjectKey, subjectAttributes, response } = configurationWire.precomputed;
+  const parsedResponse: IPrecomputedConfigurationResponse = JSON.parse(response);
```
rasendubi, line 649: Is there any reason all this parsing is done in client library and not common?

docs/js-client-sdk.iclientconfig.eventingestionconfig.md
```diff
@@ -15,5 +15,6 @@ eventIngestionConfig?: {
         maxRetryDelayMs?: number;
         maxRetries?: number;
         batchSize?: number;
+        maxQueueSize?: number;
```
rasendubi, line 18: unrelated but why do we commit auto-generated files?

src/index.ts
```diff
@@ -597,9 +589,95 @@ export async function precomputedInit(
...
+    memoryOnlyPrecomputedStore
+      .setEntries(parsedResponse.flags)
+      .catch((err) =>
+        applicationLogger.warn('Error setting precomputed assignments for memory-only store', err),
+      );
+    memoryOnlyPrecomputedStore.salt = parsedResponse.salt;
+
+    const subject: Subject = {
+      subjectKey,
+      subjectAttributes: subjectAttributes ?? ({} as Attributes),
```
rasendubi, line 662: Why do we need the cast here? is it violating type check?

src/index.ts
```diff
@@ -597,9 +589,95 @@ export async function precomputedInit(
...
+    };
+
+    EppoPrecomputedJSClient.instance = new EppoPrecomputedJSClient({
```
rasendubi, line 665: major: before potentially overwriting previous instance, we should shut it down. (and warn of re-initialization)

src/index.spec.ts
```diff
@@ -1178,6 +1139,54 @@ describe('EppoPrecomputedJSClient E2E test', () => {
...
+  beforeAll(() => {
+    precomputedConfigurationWire = readMockPrecomputedResponse(MOCK_PRECOMPUTED_WIRE_FILE);
+  });
+
+  beforeEach(() => {
+    mockLogger = td.object<IAssignmentLogger>();
+    // Reset the static instance before each test
+    jest.isolateModules(() => {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      require('./index');
```
aarsilv, line 1155: how come we need to do this instead of just force reinitializing?

src/index.spec.ts
```diff
@@ -1178,6 +1139,54 @@ describe('EppoPrecomputedJSClient E2E test', () => {
...
+      featureFlag: 'string-flag',
+      allocation: 'allocation-123',
+      variation: 'variation-123',
+      subjectAttributes: {
+        buildNumber: 42,
+        hasPushEnabled: false,
+        language: 'en-US',
+        lastLoginDays: 3,
+        lifetimeValue: 543.21,
+        platform: 'ios',
```
aarsilv, line 1178: I like that we have the different data types here!

docs/js-client-sdk.iclientconfig.eventingestionconfig.md
sameerank: we haven't gotten around to it, but I believe it's a unanimous opinion that we should remove these auto-generated docs

src/index.ts
sameerank: Nope the cast isn't needed for typescript. Did it for readability, but I'm seeing now that it's redundant. Removed

src/index.ts
sameerank: I think I'm unclear about what "shut it down" means here. The client is offline so polling/making requests isn't at play. Also the two instance variables being set by the constructor (`subject` and `precomputedFlagStore`) are private. 

I could make these public so I can clear them here. Or I can add a public method to `EppoPrecomputedClient` in the common sdk to clear these. I can't do it in the constructor because that only has access to the new instance, not the one being replaced.

src/index.ts
sameerank: Agreed and fixed!

src/index.ts
sameerank: It's due to the approach taken. The same `EppoPrecomputedClient` is used for online (via `precomputedInit`) vs offline mode (via `offlinePrecomputedInit`)

The `EppoPrecomputedClient`'s constructor knows which mode we're using based on the provided initialization options for the client instance. Both `precomputedFlagStore` and `subject` need to be accessible for initializing in the online mode: the user may want to select a specific storage option (e.g. `chrome.storage`) for their use case and `subject` is required to know what to precompute. We're reusing these same inputs for offline mode, hence the parsing.

I could add an option to pass in the configuration wire string to be parsed by the common library, which would overwrite the contents of `precomputedFlagStore`, but making `subject` optional or to overwrite it seems like doing a lot in `EppoPrecomputedClient`'s constructor.

package.json
```diff
@@ -59,7 +59,7 @@
     "webpack-cli": "^6.0.1"
   },
   "dependencies": {
-    "@eppo/js-client-sdk-common": "^4.7.1"
+    "@eppo/js-client-sdk-common": "4.8.0-alpha.0"
```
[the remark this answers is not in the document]
sameerank: This is temporary. The goal is to get everything on 4.8.0 asap. The prereleases are for passing the automated checks on github after the code updates between common library and the client sdk have been verified locally

src/index.ts
sameerank: Fixed! An error is thrown if `throwOnFailedInitialization` is true

src/index.ts
sameerank: My attempt at a private shutdown function

```
function shutdownEppoPrecomputedClient() {
  if (EppoPrecomputedJSClient.instance) {
    EppoPrecomputedJSClient.instance.stopPolling();
    EppoPrecomputedJSClient.initialized = false;
    applicationLogger.warn(`${loggerPrefix} Precomputed client is being re-initialized.`);
  }
}
```

The configuration store is a private variable in `EppoPrecomputedClient` so I can't clear it here, not does it come with any helpers to clear its contents. Also the store we use at the moment is memory-only, so I don't know if persisting across sessions is possible, e.g. if a page refresh is the reason for re-initializing

src/index.ts
rasendubi: Casts aren't a good tool for readability because they are hammers for beating types into submission—they may be hiding bugs (now or in the future if we're not careful)

Annotations don't do the hammering though and actually do a bit more of type checking, so are safe to use
```typescript
const defaultAttributes: Attributes = {};
```

src/index.ts
rasendubi: > I think I'm unclear about what "shut it down" means here.

I mostly meant stopping polling. Most SDKs have a method for stopping any activity.

> The client is offline so polling/making requests isn't at play. 

The fact that we're initializing offline doesn't mean that the previous client was in offline mode too

PR #124, opened 2024-12-11
chore: Version bump to v3.8.3

felipecsl:


sameerank (approved):

leoromanovsky (commented):
Does the current GitHub actions workflow have enough power to demonstrate that the build completes a production package?

felipecsl:
> Does the current GitHub actions workflow have enough power to demonstrate that the build completes a production package?

@leoromanovsky I don't think so, because it doens't perform any e2e tests AFAIK, only unit and integration level jest tests

felipecsl:
I will be doing that test now myself, though

PR #125, opened 2024-12-12
chore(events): Bump default batch size to 1k events

felipecsl:
---
labels: mergeable
---
Fixes: FF-3704

## Motivation and Context
Have sane defaults.

## Description
As above.

## How has this been tested?
N/A

[//]: # (OPTIONAL)


sameerank (approved):

felipecsl:
> @felipecsl maybe we should also add a minimum, say it can not be less than 100 (or 500) events?

good idea, but i'll have to add that validation in the common sdk though!

felipecsl:
@petzel do you say we just clamp it to our predefined min? so if you pass 10, we set it to 100?

PR #126, opened 2024-12-13
fix: prevent chrome storage resolving before entries are set

sameerank:
---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context

We had a hypothesis in a debugging session that `await this.storage.set` may be resolving when the writing process starts, instead of when it finishes writing the fetched flags to storage and that this may be an idiosyncratic behavior of the Chrome storage implementation. If true, this could result in inconsistent, null assignments on application start if get assignment is called immediately after `await init`

## Description

This is a draft of using an `isInProcessOfInitializing` method to wait before deciding if the store is usable

## How has this been tested?

[//]: # (OPTIONAL)

### DIFF
--- a/src/cache/chrome-storage-async-map.spec.ts
+++ b/src/cache/chrome-storage-async-map.spec.ts
@@ -90,7 +90,7 @@ describe('ChromeStorageAsyncMap', () => {
       // Mock successful storage.set
       (mockStorage.set as jest.Mock).mockResolvedValue(undefined);
 
-      // Start the set operation
+      // Create the set operation Promise
       const setPromise = storageMap.set(key, value);
 
       // Verify entry is not written yet
--- a/src/cache/chrome-storage-async-map.spec.ts
+++ b/src/cache/chrome-storage-async-map.spec.ts
@@ -37,28 +37,6 @@ describe('ChromeStorageAsyncMap', () => {
   });
 
   describe('set', () => {
-    it('should resolve when storage change event fires', async () => {
-      const key = 'testKey';
-      const value = 'testValue';
-
-      // Mock successful storage.set
-      (mockStorage.set as jest.Mock).mockResolvedValue(undefined);
-
-      // Start the set operation
-      const setPromise = storageMap.set(key, value);
-
-      // Simulate the storage change event
-      if (storedListener) {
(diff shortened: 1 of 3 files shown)


src/cache/chrome-storage-async-map.ts
```diff
@@ -18,7 +18,36 @@ export default class ChromeStorageAsyncMap<T> implements AsyncMap<string, T> {
...
+        };
+
+        chrome.storage.onChanged.addListener(listener);
+
+        // Set a timeout in case the change event never fires
+        const timeout = setTimeout(() => {
+          chrome.storage.onChanged.removeListener(listener);
+          applicationLogger.warn('Chrome storage write timeout for key:', key);
+          resolve();
+        }, 1000);
```
sameerank, line 39: One second is probably too long, but we also never expect to end up in this situation of needing the timeout. It's a precaution

src/cache/chrome-storage-async-map.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+          return {}; // Entry not yet written
+        }
+        return { [key]: value }; // Entry written after change event
+      });
+
+      // Mock successful storage.set
+      (mockStorage.set as jest.Mock).mockResolvedValue(undefined);
+
+      // Start the set operation
+      const setPromise = storageMap.set(key, value);
```
sameerank, line 94: This isn't exactly what I am hoping to test since the set operation isn't in progress, and I'm interested in testing that an entry hasn't been written yet between the time that the operation started and onChange was fired. I couldn't figure out how to write such a test.

src/string-valued.store.ts
```diff
@@ -19,13 +19,18 @@ export interface IStringStorageEngine {
...
   private initialized = false;
+  private isInitializing = false;
 
   public constructor(private storageEngine: IStringStorageEngine, private cooldownSeconds = 0) {}
 
   public isInitialized(): boolean {
     return this.initialized;
   }
 
+  public isInProcessOfInitializing(): boolean {
```
sameerank, line 30: TODO add `isInProcessOfInitializing` to the `IAsyncStore` interface

PR #131, opened 2024-12-16
feat: implement overrides store functionality

yfrancis:
Add a new config parameter `enableOverrides`, which when set will create an overrides store and supply it to the eppo client on init. The overrides store leverages a hybrid store with memory and window localStorage for persistence.

Depends on https://github.com/Eppo-exp/js-sdk-common/pull/184


aarsilv (commented):
This is a great start! I think we'll want a test in `index.spec` that spins up an instance, applies an override, and retrieves a value (among other things) to be sure the integration between this SDK and the common one works as expected.

greghuels (approved):

PR #135, opened 2025-01-02
Bump common version to allow handling of special characters

aarsilv:
_Eppo Internal:_
🎟️ **Ticket:** [FF-3767 - Update JS Client to use updated Common JS for special character handling](https://linear.app/eppo/issue/FF-3767/update-js-client-to-use-updated-common-js-for-special-character)
🗨️ **Slack Thread:** ["...flag text encoding problem..."](https://eppo-group.slack.com/archives/C0541NTN6QH/p1735566355776199)
👯 **Related PR:** [`js-sdk-common #187`](https://github.com/Eppo-exp/js-sdk-common/pull/187)

## Motivation and Context
SDKs need to be able to handle special, non-ASCII characters. This PR adds a test that checks that by using strings with characters from a variety of languages as well as emojis, which all have non-ASCII values.

## Description
We update our common module which switches from using [`js-base64`](https://www.npmjs.com/package/js-base64)'s `btoaPolyfill()` and `atobPolyfill()` to `encode` and `decode` respectively.

The former is designed only to handle ASCII characters and bytes ([docs](https://github.com/dankogai/js-base64?tab=readme-ov-file#decode-vs-atob-and-encode-vs-btoa)), while the latter is designed for strings and can handle all characters. 

Also in here are some linter-suggested fixes

## How has this been tested?
[Updated shared test case](https://github.com/Eppo-exp/sdk-test-data/pull/103)

### DIFF
--- a/src/events/local-storage-backed-named-event-queue.ts
+++ b/src/events/local-storage-backed-named-event-queue.ts
@@ -47,7 +47,10 @@ export default class LocalStorageBackedNamedEventQueue<T> implements NamedEventQ
     if (this.eventKeys.length === 0) {
       return undefined;
     }
-    const eventKey = this.eventKeys.shift()!;
+    const eventKey = this.eventKeys.shift();
+    if (eventKey === undefined) {
+      throw new Error('Unexpected undefined event key');
+    }
     const eventData = localStorage.getItem(eventKey);
     if (eventData) {
       localStorage.removeItem(eventKey);
--- a/src/events/local-storage-backed-named-event-queue.ts
+++ b/src/events/local-storage-backed-named-event-queue.ts
@@ -48,7 +48,7 @@ export default class LocalStorageBackedNamedEventQueue<T> implements NamedEventQ
       return undefined;
     }
     const eventKey = this.eventKeys.shift();
-    if (eventKey === undefined) {
+    if (!eventKey) {
       throw new Error('Unexpected undefined event key');
     }
     const eventData = localStorage.getItem(eventKey);
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -1188,8 +1188,8 @@ describe('EppoClient config', () => {
         json: () => Promise.resolve({}),
       });
     }) as jest.Mock;
-    EppoJSClient.initialized = false;
     const client = await init({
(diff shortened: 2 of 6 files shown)


src/index.spec.ts
```diff
@@ -1188,24 +1188,24 @@ describe('EppoClient config', () => {
...
+      forceReinitialize: true,
       apiKey: 'zCsQuoHJxVPp895.ZWg9MTIzNDU2LmUudGVzdGluZy5lcHBvLmNsb3Vk',
       assignmentLogger: td.object<IAssignmentLogger>(),
       eventIngestionConfig: {
         deliveryIntervalMs: 1,
         retryIntervalMs: 2,
         maxRetryDelayMs: 3,
         maxRetries: 4,
-        batchSize: 5,
+        batchSize: 500,
```
aarsilv, line 1199: Had to update as the common module now clamps between 100 and 1,000

src/index.spec.ts
```diff
@@ -1188,24 +1187,24 @@ describe('EppoClient config', () => {
         json: () => Promise.resolve({}),
       });
     }) as jest.Mock;
-    EppoJSClient.initialized = false;
     const client = await init({
+      forceReinitialize: true,
```
aarsilv, line 1191: The "official" way to ensure a new instance is spun up (vs. flipping the boolean)

src/events/local-storage-backed-named-event-queue.ts
```diff
@@ -47,7 +47,10 @@ export default class LocalStorageBackedNamedEventQueue<T> implements NamedEventQ
     if (this.eventKeys.length === 0) {
       return undefined;
     }
-    const eventKey = this.eventKeys.shift()!;
+    const eventKey = this.eventKeys.shift();
+    if (eventKey === undefined) {
```
felipecsl, line 51: why not simply `if (!eventKey)`?

felipecsl (approved):
lgtm, just one question

src/events/local-storage-backed-named-event-queue.ts
aarsilv: Yeah good call, this was an autofix by the IDE 😅

PR #137, opened 2025-01-02
feat(events): Allow providing a `maxQueueSize` configuration

felipecsl:
## Motivation and Context
Allow providing a `maxSize` configuration for `BoundedEventQueue`.

## Description
This code path is only hit when `localStorage` is not available, which should be the "unhappy" path. In that case, we want to allow keeping more than 100 events in memory (`MAX_EVENT_QUEUE_SIZE`)

## How has this been tested?
Didn't write a test since it's trivial, but I can if folks feel like it's important enough

[//]: # (OPTIONAL)


aarsilv (approved):
👍

PR #143, opened 2025-01-10
Typo/fix/race from old ver ff 3708

typotter:
---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context

## Description

## How has this been tested?

[//]: # (OPTIONAL)

### DIFF
--- a/src/i-client-config.ts
+++ b/src/i-client-config.ts
@@ -2,6 +2,8 @@ import { AttributeType, Flag, IAssignmentLogger, IAsyncStore } from '@eppo/js-cl
 
 import { ServingStoreUpdateStrategy } from './isolatable-hybrid.store';
 
+export const DEFAULT_LOCAL_STORE_HEAD_START_MS = 10;
+
 /**
  * Base configuration for API requests and polling behavior
  */
@@ -26,6 +28,11 @@ interface IBaseRequestConfig {
    * Timeout in milliseconds for the HTTPS request for the experiment configuration. (Default: 5000)
    */
   requestTimeoutMs?: number;
+
+  /**
+   * Time to wait, in milliseconds for the persistent store to initialize before polling on initialization. (Default: 10)
+   */
+  localStoreHeadstartMs?: number;
 
   /**
    * Number of additional times the initial configuration request will be attempted if it fails.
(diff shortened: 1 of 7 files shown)


src/i-client-config.ts
```diff
@@ -2,6 +2,8 @@ import { AttributeType, Flag, IAssignmentLogger, IAsyncStore } from '@eppo/js-cl
 
 import { ServingStoreUpdateStrategy } from './isolatable-hybrid.store';
 
+export const DEFAULT_LOCAL_STORE_HEAD_START_MS = 10;
```
aarsilv, line 5: I wonder if we want a more descriptive variable name, although I can't think of a good one.

`DEFAULT_PERSISENT_STORE_INIT_WINDOW`?

Either way I think we use the "persistent" messaging more than "local"

src/index.ts
```diff
@@ -409,25 +413,30 @@ export async function init(config: IClientConfig): Promise<EppoClient> {
     instance.setEventDispatcher(newEventDispatcher(apiKey, eventIngestionConfig));
 
     // We have two at-bats for initialization: from the configuration store and from fetching
+    // The remote fetch races against a local persistent store, however, the
```
aarsilv, line 416: Half finished comment!

https://www.youtube.com/watch?v=ZlIz0q8aWpA

src/index.ts
```diff
@@ -437,22 +446,41 @@ export async function init(config: IClientConfig): Promise<EppoClient> {
...
+        return 'did not finish';
       });
 
+    const persistentStoreHeadStartMs =
+      config.localStoreHeadstartMs ?? DEFAULT_LOCAL_STORE_HEAD_START_MS;
+    const headStart = new Promise<ConfigSources>((resolve) => {
+      setTimeout(() => resolve('head start'), persistentStoreHeadStartMs);
+    });
+
+    const headStartWinner = await Promise.race([headStart, attemptInitFromConfigStore]);
```
aarsilv, line 468: ~I reckon we could still call this "initializationSource" no need to bring headStartIntoThis~

OK I see what's happening, I think we'll want more code comments because it's a bit confusing.

Basically we want the config store to race against a 10ms timeout, but if the timeout wins, we allow it to keep going in a new race against fetch. If it wins, we don't bother with fetch at all.

src/index.ts
```diff
@@ -461,7 +489,7 @@ export async function init(config: IClientConfig): Promise<EppoClient> {
       }
     }
 
-    if (!initializationSource) {
+    if (initializationSource === 'did not finish') {
```
aarsilv, line 492: This is more clear good thinking. Given how much we use these in the code repeatedly, perhaps we should make it an enum or something.

PR #145, opened 2025-01-13
fix: Multiple "concurrent" calls to `init` causing multiple fetches.

typotter:
---
labels: mergeable
---
Fixes: #FF-3708

## Motivation and Context
Eppo saw a dramatic increase in requests to the API and it was determined through investigation that multiple calls to `init` were being made before previous calls finished. Our code did not handle that case and would run the initialization workload (including fetching) for each `init` call until one of them resolved. This is bad because it causes additional network and CPU load for the client and increases the cost of delivery for Eppo.

## Description
- Map of API Key => Promise<EppoClient>
- init-wrapping method (`init` renamed to `exclusiveInit` and new method named `init` to preserve API).
- check the map for a value at the start of `init` and delegate if one exists.
- delete the value from the map when init completes - this indicates that the initialization workload is complete

## How has this been tested?
- new tests

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -503,6 +503,40 @@ describe('initialization options', () => {
     expect(callCount).toBe(2);
   });
 
+  it('Multiple Inits', async () => {
+    let callCount = 0;
+
+    global.fetch = jest.fn(() => {
+      ++callCount;
+      return Promise.resolve({
+        ok: true,
+        status: 200,
+        json: () => Promise.resolve(mockConfigResponse),
+      });
+    }) as jest.Mock;
+
+    const inits: Promise<EppoClient>[] = [];
+    [...Array(10).keys()].forEach(() => {
+      inits.push(
+        init({
+          apiKey,
+          baseUrl,
+          assignmentLogger: mockLogger,
+        }),
+      );
+    });
+
+    // Advance timers mid-init to allow retrying
+    await jest.advanceTimersByTimeAsync(maxRetryDelay);
+
+    // Await for all the initialization calls to resolve
+    const client = await Promise.race(inits);
+    await Promise.all(inits);
+
+    expect(callCount).toBe(1);
+    expect(client.getStringAssignment(flagKey, 'subject', {}, 'default-value')).toBe('control');
+  });
+
   it('do not reinitialize if already initialized', async () => {
     let callCount = 0;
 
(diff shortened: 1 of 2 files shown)


src/index.spec.ts
```diff
@@ -503,6 +503,104 @@ describe('initialization options', () => {
...
+
+    // Advance timers mid-init to allow retrying
+    await jest.advanceTimersByTimeAsync(maxRetryDelay);
+
+    // Await for all the initialization calls to resolve
+    const client = await Promise.race(inits);
+    await Promise.all(inits);
+
+    expect(callCount).toBe(1);
+    expect(client.getStringAssignment(flagKey, 'subject', {}, 'default-value')).toBe('control');
```
felipecsl, line 537: nice test

src/index.ts
```diff
@@ -315,6 +315,13 @@ export function offlineInit(config: IClientConfigSync): EppoClient {
   return EppoJSClient.instance;
 }
 
+type APIKey = string;
```
felipecsl, line 318: nit: `SDKKey` is more accurate to avoid confusion

src/index.ts
```diff
@@ -323,6 +330,24 @@ export function offlineInit(config: IClientConfigSync): EppoClient {
...
+    return _initPromise;
+  }
+
+  _initPromise = explicitInit(config);
+  _initializationPromises.set(config.apiKey, _initPromise);
+
+  return _initPromise.then((client) => {
+    _initializationPromises.delete(config.apiKey);
+    return client;
+  });
```
felipecsl, line 346: since this is an `async function`, you can just `await _initPromise` instead of using `then`, that's a bit cleaner

src/index.ts
```diff
@@ -323,6 +330,24 @@ export function offlineInit(config: IClientConfigSync): EppoClient {
...
+}
```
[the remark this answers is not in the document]
typotter: 😳 not enough coffee to remember I was writing javascript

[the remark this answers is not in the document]
felipecsl: 🤣

src/index.ts
typotter: agreed

src/index.ts
typotter: done

typotter:
Thanks for the review, Felipe. ptal

felipecsl (approved):
🚢

PR #149, opened 2025-01-16
fix: allow empty string for precomputedConfiguration

sameerank:
---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context

## Description

We noticed an uncaught exception when `precomputedConfiguration` is an empty string. To be consistent with the online `precomputedInit`, this function should also always return an instance if `throwOnFailedInitialization` is `false`

## How has this been tested?

[//]: # (OPTIONAL)

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -1342,4 +1342,25 @@ describe('EppoClient config', () => {
     expect(retryManager['config']['maxRetryDelayMs']).toEqual(3);
     expect(retryManager['config']['maxRetries']).toEqual(4);
   });
+
+  it('handles empty precomputed configuration string', () => {
+    // Test with throwOnFailedInitialization = true (default)
+    expect(() =>
+      offlinePrecomputedInit({
+        precomputedConfiguration: '',
+      }),
+    ).toThrow('Invalid precomputed configuration wire');
+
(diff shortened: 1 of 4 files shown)


src/index.spec.ts
```diff
@@ -1342,4 +1342,25 @@ describe('EppoClient config', () => {
     expect(retryManager['config']['maxRetryDelayMs']).toEqual(3);
     expect(retryManager['config']['maxRetries']).toEqual(4);
   });
+
+  it('handles empty precomputed configuration string', () => {
+    // Test with throwOnFailedInitialization = true (default)
+    expect(() =>
+      offlinePrecomputedInit({
+        precomputedConfiguration: '',
```
leoromanovsky, line 1350: is the exception specifically an empty string or any malformed configuration json?

leoromanovsky (approved):

src/index.spec.ts
sameerank: The exception specifically in mind was an empty string which seems likely to be a fallback in loading the configuration string, i.e.

```
let configuration = '';
... and then set configuration to precomputed string
```

but this would also catch any other string that cannot be parsed to json, which is useful too. The precomputed configuration should always be valid json

PR #158, opened 2025-01-21
fix: solve race condition resulting in empty, uninitialized configuration store.

typotter:
_Eppo Internal_
Fixes: FF-3708

## Motivation and Context
A race condition existed whereby the reading entries and metadata from the persistent store could invert resulting in a "ready" EppoClient which had an uninitialized (and empty) configuration store.
When a client is initialized and immediately used for assignment, it will sometimes return the default instead of a computed assignment. This was eventually determined to be caused by a set of race conditions affecting the initialization flow.

## Description
- First, we enumerate the possible status of a configuration load attempt (failed, incomplete, complete) and config sources `ConfigSource`
- Next, we more explicitly track which attempt finishes first and its status. Similarly for the second attempt if the first did not complete.
- Check to see if the configuration store `isInitialized()` when the `fetchConfiguration` method returns. This verifies whether the fetch _actually set any data in the configuration store_ and is the only signal we have at this level that the fetch logic either fetched and stored config, encountered an error, got an invalid config response, _or gave up on the fetch because the config store was not yet expired_

## How has this been tested?
- new test case exercising the race condition
- painstaking debugging of other tests to understand all the fine details of the config loading flow and its dependencies


src/index.spec.ts
```diff
@@ -538,7 +538,7 @@ describe('initialization options', () => {
     expect(client.getStringAssignment(flagKey, 'subject', {}, 'default-value')).toBe('control');
   });
 
-  it('only fetches/does initialization workload once per API key if init is called multiple times concurrently', async () => {
+  it.skip('only fetches/does initialization workload once per API key if init is called multiple times concurrently', async () => {
```
typotter, line 541: So, this test fails, not because of the change made in `index.ts` to only buffer one initialization attempt. It fails because the SDK does not actually support instantiating and using multiple clients with different SDKs. The initialization method does all the work on the EppoJsClient.instance singleton. Thus, when init is called with a second SDK/API Key, it overwrites the config store and other properties of the instance loading the first key. By the time all the inits in this test are queued, the reference to `configurationStore` [here](https://github.com/Eppo-exp/js-client-sdk/pull/158/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R502) gets out of sync with `instance.configurationStore` for all but the last SDK key to resolve

aarsilv: Ah I see the issue is with multiple _different_ API keys. I agree this is currently unsupported functionality. Consider either removing this or adding a comment on how it's commented out because we don't support this... YET

src/index.spec.ts
```diff
@@ -1137,6 +1141,69 @@ describe('initialization options', () => {
...
+      // Complete the "load from cache"
+      if (entriesPromise.resolve) {
+        entriesPromise.resolve(mockConfigResponse.flags);
+      } else {
+        throw 'Error running test';
+      }
+
+      // Await so it can finish its initialization before this test proceeds
+      const client = await initPromise;
+      expect(callCount).toBe(0);
```
aarsilv, line 1203: 👍

src/index.spec.ts
```diff
@@ -1378,3 +1445,20 @@ describe('EppoClient config', () => {
     });
   });
 });
+
+/**
+ * A wrapper for a promise which allows for later resolution.
+ */
+class DeferredPromise<T> {
```
aarsilv, line 1452: ❗ woah neato!

Cool concept this would have cleaned up many of the other tests in here had I thought of this 🤦

src/index.ts
```diff
@@ -350,17 +349,15 @@ export async function init(config: IClientConfig): Promise<EppoClient> {
   validation.validateNotBlank(config.apiKey, 'API key required');
 
   // If there is already an init in progress for this apiKey, return that.
-  let initPromise = initializationPromises.get(config.apiKey);
-  if (initPromise) {
-    return initPromise;
+  // let initPromise = initializationPromises.get(config.apiKey);
```
aarsilv, line 352: shall we remove this line?

src/index.ts
```diff
@@ -459,60 +456,103 @@ async function explicitInit(config: IClientConfig): Promise<EppoClient> {
...
-    const attemptInitFromConfigStore = configurationStore
+    const attemptInitFromConfigStore: ConfigurationLoadAttempt = configurationStore
       .init()
       .then(async () => {
         if (!configurationStore.getKeys().length) {
           // Consider empty configuration stores invalid
           applicationLogger.warn('Eppo SDK cached configuration is empty');
           initFromConfigStoreError = new Error('Configuration store was empty');
-          return '';
+          return ConfigLoaderStatus.DID_NOT_PRODUCE;
```
aarsilv, line 474: enums making this much cleaner 👌

src/config-loader.ts
```diff
@@ -0,0 +1,16 @@
...
+  COMPLETED, // Useful configuration was successfully loaded into the configuration store from this source.
+}
+
+export enum ConfigSource {
+  CONFIG_STORE,
+  FETCH,
+  NONE,
+}
+
+export type ConfigurationLoadAttempt = Promise<[ConfigSource, ConfigLoaderStatus]>;
```
aarsilv, line 16: Typically, I'd expect to see this as an object (as opposed to an array) as it would make accessing the two fields a bit more explicit, but I don't have a strong preference either way

src/index.ts
```diff
@@ -459,60 +456,103 @@ async function explicitInit(config: IClientConfig): Promise<EppoClient> {
...
+      initializationSource = await secondAttempt.then(([source, result]) => {
+        return result === ConfigLoaderStatus.COMPLETED ? source : ConfigSource.NONE;
+      });
     }
 
-    if (!initializationSource) {
-      // both failed, make the "fatal" error the fetch one
-      initializationError = initFromFetchError;
+    if (initializationSource === ConfigSource.NONE) {
+      console.log(initFromConfigStoreError);
```
aarsilv, line 549: `applicationLogger.warn()`?

src/index.ts
```diff
@@ -459,60 +456,103 @@ async function explicitInit(config: IClientConfig): Promise<EppoClient> {
...
+      .then((status) => {
+        return [ConfigSource.CONFIG_STORE, status];
       });
-    const attemptInitFromFetch = instance
+
+    const attemptInitFromFetch: ConfigurationLoadAttempt = instance
       .fetchFlagConfigurations()
-      .then(() => 'fetch')
+      .then(() => {
+        console.log(instance.getFlagConfigurations());
```
aarsilv, line 503: probably want to clean this up

src/index.ts
```diff
@@ -459,60 +456,103 @@ async function explicitInit(config: IClientConfig): Promise<EppoClient> {
...
+        if (configurationStore.isInitialized()) {
+          // If fetch has won the race and the configStore is initialized, that means we have a successful load and the
+          // fetched data was used to init.
+          return ConfigLoaderStatus.COMPLETED;
+        } else {
+          // If the config store is not initialized and the fetch says it is done, that means the fetch did not set any
+          // values into the config store (ex:the cache is not expired or invalid config data was received).
```
aarsilv, line 510: If the cache is not expired then it would be initialized wouldn't it? I think did not produce means it was invalid or empty

src/index.ts
```diff
@@ -459,60 +456,103 @@ async function explicitInit(config: IClientConfig): Promise<EppoClient> {
...
-    if (!initializationSource) {
-      // First attempt failed, but we have a second at bat that will be executed in the scope of the top-level try-catch
-      if (!initFromConfigStoreError) {
-        initializationSource = await attemptInitFromConfigStore;
-      } else {
-        initializationSource = await attemptInitFromFetch;
-      }
+    // attemptInitFromFetch only has side effects if `skipInitialRequest` is false.
+    if (!config.skipInitialRequest) {
+      racers.push(attemptInitFromFetch);
```
aarsilv, line 529: No more slightly delayed start for fetch? How do we prevent fetch from trying if the configuration store inits very quickly? Or is this the new paradigm where by the time we get here everything should be initialized? (sorry--hard for me to keep track of all the conversation 😛 )

aarsilv (approved):
Nice job sussing this all out!

Approving assuming fetches won't be made if a valid, unexpired cache exists. I _think_ this is true based on the automated test, but it is hard for me to see what is stopping that at the code. Is it because we now initialize our factory-created stores off the bat?

src/index.ts
typotter: thrown/logged later; removed

src/index.ts
typotter: The poller callback does a check to see if the config is expired before making the actual network call. This happens in the common lib. Added some more comments here and elsewhere.

PR #164, opened 2025-01-28
feat: assignment cache to dedupe logging by the precomputed client

sameerank:
---
labels: mergeable
---
Fixes: #__issue__

## Motivation and Context

The precompute client needs the assignment cache to dedupe assignments. This is central functionality of the SDKs, not a nice-to-have

## Description

Scope of work:

- [x] offline init - use in-memory assignment cache
- [x] online init - use hybrid assignment cache

## How has this been tested?

[//]: # (OPTIONAL)

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -1316,10 +1316,7 @@ describe('EppoPrecomputedJSClient E2E test', () => {
   it('deduplicates assignment logging', () => {
     // Reset the mock logger and assignment cache before this test
     const mockLogger = td.object<IAssignmentLogger>();
-    const mockAssignmentCache = td.object<AssignmentCache>();
-    td.when(mockAssignmentCache.has(td.matchers.anything())).thenReturn(false, true);
-    td.when(mockAssignmentCache.set(td.matchers.anything())).thenReturn();
-    globalClient.useCustomAssignmentCache(mockAssignmentCache);
+    globalClient.useNonExpiringInMemoryAssignmentCache();
     globalClient.setAssignmentLogger(mockLogger);
 
     expect(td.explain(mockLogger.logAssignment).callCount).toEqual(0);
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -1282,6 +1282,8 @@ describe('EppoPrecomputedJSClient E2E test', () => {
   });
 
   it('logs assignments correctly', () => {
+    const mockAssignmentCache = td.object<AssignmentCache>();
+    globalClient.useCustomAssignmentCache(mockAssignmentCache);
     // Reset the mock logger before this test
     mockLogger = td.object<IAssignmentLogger>();
     globalClient.setAssignmentLogger(mockLogger);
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -694,6 +694,16 @@ export async function precomputedInit(
...
 
+  // Add assignment cache initialization
+  const storageKeySuffix = buildStorageKeySuffix(apiKey);
+  const assignmentCache = assignmentCacheFactory({
+    chromeStorage: chromeStorageIfAvailable(),
+    storageKeySuffix,
+  });
+  if (assignmentCache instanceof HybridAssignmentCache) {
+    await assignmentCache.init();
+  }
```
sameerank, line 705: Uses a hybrid assignment cache for the fetch-based `precomputedInit` method

src/index.ts
```diff
@@ -804,6 +815,13 @@ export function offlinePrecomputedInit(
     };
 
     shutdownEppoPrecomputedClient();
+
+    // Add memory-only assignment cache for offline mode
+    const assignmentCache = assignmentCacheFactory({
+      storageKeySuffix: 'offline',
+      forceMemoryOnly: true,
+    });
```
sameerank, line 823: Uses an in-memory assignment cache for offline initialization

src/index.spec.ts
```diff
@@ -1310,6 +1312,27 @@ describe('EppoPrecomputedJSClient E2E test', () => {
       format: 'PRECOMPUTED',
     });
   });
+
+  it('deduplicates assignment logging', () => {
+    // Reset the mock logger and assignment cache before this test
+    const mockLogger = td.object<IAssignmentLogger>();
+    const mockAssignmentCache = td.object<AssignmentCache>();
+    td.when(mockAssignmentCache.has(td.matchers.anything())).thenReturn(false, true);
```
sameerank, line 1320: I needed to reset the assignment cache at the top of this test, and mocking its behavior is one way to do it. I don't like this approach though because telling the function when to return false or true is removing the part of the test that ensures that the assignment cache is setting assignments correctly

Some alternative approaches

Use the same code as the precomputedInit method to initialize a new hybrid cache

https://github.com/Eppo-exp/js-client-sdk/blob/56c4555e7a2ac1b11257eb6297e79c9eca4a1c0a/src/index.ts#L697-L705

Or switch to using an in-memory cache. The methods for enabling this are convenient

```
globalClient.useNonExpiringInMemoryAssignmentCache()
```

src/index.spec.ts
```diff
@@ -1282,6 +1282,8 @@ describe('EppoPrecomputedJSClient E2E test', () => {
   });
 
   it('logs assignments correctly', () => {
+    const mockAssignmentCache = td.object<AssignmentCache>();
+    globalClient.useCustomAssignmentCache(mockAssignmentCache);
```
sameerank, line 1286: Another case of needing to reset the assignment cache (filled by other tests running) to get logging to work correctly in this test specifically

src/index.ts
felipecsl: why do we need to check on the type of `assignmentCache` here? I'd prefer if we didn't make any assumptions on the actual implementation. Can that `init()` call be handled inside the factory instead?

sameerank: This is copying a pattern that is already in place for the non-precomputed init methods, so I don't know the original intent. One benefit I see from looking at the code though is that it's here to keep the `await` out of the `assignmentCacheFactory`, which ensures that the factory method is synchronous and therefore can be used in both the online and offline init methods

aarsilv: yeah this seems like necessary evil around mixing async and sync code.

I suppose another thing you could do is check to see if the cache has an init() method, and if so await init it. A little more future proof perhaps.

src/index.ts
aarsilv: How come? Local storage or chome storage could still be "offline"?

aarsilv (approved):
Thanks for thinking this through! Approving with some musings on the assignment cache you can take or leave as you see fit!

src/index.ts
sameerank: In-memory is consistent with how we do it in the `offlineInit` method

https://github.com/Eppo-exp/js-client-sdk/blob/dced1c2ded84a2173e6394542b16dd700da6931e/src/index.ts#L320-L327

I think it's a persistence vs access speed trade-off. The hybrid cache bridges the two goals, but won't work in the sync function due to the `assignmentCache.init` method being async. By choosing to go with an in-memory cache, we prioritized application speed over persistence across browser sessions and tabs, which makes sense since speed is likely also the priority of an SDK user who opts for an offline-init pattern

leoromanovsky: > Local storage or chome storage

chrome storage is async and cannot be used here. although local storage is sync, it is so slow that I have been only wrappng it in async calls and I don't think its a good store to put into the execution path for evaluation.

PR #166, opened 2025-01-31
feat: buildAndInit for isolated client instances

typotter:
---
labels: mergeable
---
_Eppo Internal_
🎟️ Fixes: FF-3888
📜 [Multiple Eppo Clients](https://www.notion.so/eppo/Multiple-concurrent-and-isolated-EppoClient-instances-running-within-an-application-18449cc0114380288eb3eaac580c4e53)
[Configuration Side-loading](https://www.notion.so/eppo/Configuration-side-loading-design-12249cc0114380bc9650d8dc4e189e77)

## Motivation and Context
The `init` and `getInstance()` methods work on a singleton instance of the `EppoClient` which makes for a generally smooth developer experience given that the Eppo SDK essentially handles dependency management for the dev. This does have its own drawbacks, but those are not particularly relevant here.

There are use cases, however, when a non-singleton instance of the `EppoClient` is required. One such use case is embedding the Eppo SDK into a library which itself can be included in other applications. If that 3p library shipped with Eppo is integrated into another application that has the Eppo SDK running, there would be undefined behaviour as the SDK cannot handle this use case.

The `init` method (and class constructors) for `EppoClient` and subclasses have evolved organically over time, adding new options as new features are added to the clients. The very large options type is beginning to become a little untenable and disorganized so we take the opportunity to clean that up a bit here.

There are other limitations and drawbacks to the current model of instantiating an `EppoClient` statically and then initializing it when the code calls `init` including the awkward coupling of needing to wait for init to resolve in order to get a reference to an initialized client. We have an opportunity to decouple initialization and waiting to make for a better DX (in addition to giving the dev full control over managing the `EppoClient` reference (intrinsically allows for easier mocking in tests and will be consistent with the host applications existing DI approach).

**This change must be done without a major version bump and completely preserve the existing singleton API**

## Description
- Refactor the initialization options by grouping related options and extracting different types for each.
- Each option type is then combined back into a type using the existing `IClientConfig` name (this keeps the change backwards compatible while offering a big win in option clarity)
**- New method: `buildAndInit`** which creates an isolated client instance.
**- new method: `waitForConfiguration`** which resolves when the first config is loaded.
- Docs

- https://linear.app/eppo/issue/FF-3926/[docs]-update-js-client-docs-to-describe-the-non-singleton-client

## How has this been tested?
- tests


package.json
```diff
@@ -60,7 +60,8 @@
     "webpack-cli": "^6.0.1"
   },
   "dependencies": {
-    "@eppo/js-client-sdk-common": "4.8.4"
+    "@eppo/js-client-sdk-common": "4.8.4",
+    "spark-md5": "^3.0.2"
```
typotter, line 64: Temp workaround until [js-common#207](https://github.com/Eppo-exp/js-sdk-common/pull/207) lands and is released

src/i-client-config.ts
```diff
@@ -104,10 +105,41 @@ export interface IPrecomputedClientConfig extends IBaseRequestConfig {
...
- * Configuration for regular client initialization
- * @public
+ * Base options for the EppoClient SDK
  */
-export interface IClientConfig extends IBaseRequestConfig {
+export type IApiOptions = {
+  /**
+   * Your key for accessing Eppo through the Eppo SDK.
+   */
+  sdkKey: string;
```
typotter, line 114: change from current `apiKey`. newly named `sdkKey` is used to differentiate configuration structs passed to the EppoJSClient constructor.
the type `IClientConfig` omits this field and uses `apiKey` instead to preserve backwards compatibility.

src/index.ts
```diff
@@ -98,22 +121,75 @@ const memoryOnlyPrecomputedBanditsStore = precomputedBanditStoreFactory();
...
+   * @private
+   */
+  private readyPromiseResolver: (() => void) | null = null;
+
   public static instance = new EppoJSClient({
     flagConfigurationStore,
     isObfuscated: true,
   });
-  public static initialized = false;
+  initialized = false;
```
typotter, line 141: move static members to the instance

src/index.ts
```diff
@@ -98,22 +121,75 @@ const memoryOnlyPrecomputedBanditsStore = precomputedBanditStoreFactory();
...
+      this.readyPromise = explicitInit(
+        convertClientOptionsToClientConfig(optionsOrConfig),
+        this,
+      ).then(() => {
+        return;
+      });
+    }
+  }
+
+  public waitForReady(): Promise<void> {
```
typotter, line 179: new method! 🎉

src/index.ts
```diff
@@ -98,22 +121,72 @@ const memoryOnlyPrecomputedBanditsStore = precomputedBanditStoreFactory();
...
```
typotter, line 179: new method! 🎉

src/index.ts
```diff
@@ -98,22 +121,72 @@ const memoryOnlyPrecomputedBanditsStore = precomputedBanditStoreFactory();
...
+  }
 
   public getStringAssignment(
     flagKey: string,
     subjectKey: string,
     subjectAttributes: Record<string, AttributeType>,
     defaultValue: string,
   ): string {
-    EppoJSClient.ensureInitialized();
+    this.ensureInitialized();
```
typotter, line 189: moving static methods and fields to the instance

src/index.ts
```diff
@@ -252,26 +325,58 @@ export class EppoJSClient extends EppoClient {
...
     return super.getExperimentContainerEntry(flagExperiment, subjectKey, subjectAttributes);
   }
 
-  private static ensureInitialized() {
-    if (!EppoJSClient.initialized) {
+  private ensureInitialized() {
+    if (!this.initialized) {
+      // TODO: check super.isInitialized?
```
typotter, line 334: `EppoClient.initialized` also exists, but means something different than it does here.

src/index.ts
```diff
@@ -252,26 +325,58 @@ export class EppoJSClient extends EppoClient {
...
+        return client;
+      });
+    }
+
+    const readyClient = await EppoJSClient.initializationPromise;
+    EppoJSClient.initializationPromise = null;
+
+    return readyClient;
+  }
 }
```
typotter, line 380: This is the old `init` method and initialization buffer tracker.
Moved the method to a static member of `EppoJSClient` so that it can access the `readyResolver` and notify `waitForReady`

src/index.ts
```diff
@@ -287,6 +392,8 @@ export function buildStorageKeySuffix(apiKey: string): string {
  * @public
  */
 export function offlineInit(config: IClientConfigSync): EppoClient {
+  const instance = getInstance();
```
typotter, line 395: prep for future refactoring (i.e. offline init non-singleton client)

src/index.ts
```diff
@@ -335,41 +442,31 @@ export function offlineInit(config: IClientConfigSync): EppoClient {
     }
   }
 
-  EppoJSClient.initialized = true;
-  return EppoJSClient.instance;
+  instance.initialized = true;
+  return instance;
 }
 
-/**
```
typotter, line 342: moved to `EppoJSClient`

src/index.ts
```diff
@@ -335,41 +442,31 @@ export function offlineInit(config: IClientConfigSync): EppoClient {
...
-
 /**
  * Initializes the Eppo client with configuration parameters.
  * This method should be called once on application startup.
  * If an initialization is in process, calling `init` will return the in-progress
  * `Promise<EppoClient>`. Once the initialization completes, calling `init` again will kick off the
  * initialization routine (if `forceReinitialization` is `true`).
+ *
+ *
+ * @deprecated
```
typotter, line 457: controversial, I know.
willing to negotiate

src/index.ts
```diff
@@ -335,41 +442,31 @@ export function offlineInit(config: IClientConfigSync): EppoClient {
...
+ * Use `new EppoJSClient(options)` instead of `init` or `initializeClient`. These will be removed in v4
+ *
  * @param config - client configuration
  * @public
  */
 export async function init(config: IClientConfig): Promise<EppoClient> {
-  validation.validateNotBlank(config.apiKey, 'API key required');
```
typotter, line 357: moved to `EppoJSClient`

src/index.ts
```diff
@@ -335,41 +442,31 @@ export function offlineInit(config: IClientConfigSync): EppoClient {
...
-  }
-
-  const client = await initializationPromise;
-  initializationPromise = null;
-  return client;
+  return EppoJSClient.initializeClient(getInstance(), config);
 }
 
-async function explicitInit(config: IClientConfig): Promise<EppoClient> {
+async function explicitInit(config: IClientConfig, instance: EppoJSClient): Promise<EppoJSClient> {
```
typotter, line 467: instance param instead of operating on the singleton

src/index.ts
```diff
@@ -597,7 +696,7 @@ async function explicitInit(config: IClientConfig): Promise<EppoClient> {
  * @returns a singleton client instance
  * @public
  */
-export function getInstance(): EppoClient {
+export function getInstance(): EppoJSClient {
```
typotter, line 699: More precise typing

src/index.ts
rasendubi, line 467: minor: wonder if we can make it an instance method now that it accepts an instance?

src/index.ts
```diff
@@ -252,26 +325,58 @@ export class EppoJSClient extends EppoClient {
...
+   * @param client
+   * @param config
+   */
+  static async initializeClient(
+    client: EppoJSClient,
+    config: IClientConfig,
+  ): Promise<EppoJSClient> {
+    validation.validateNotBlank(config.apiKey, 'API key required');
+
+    // If there is already an init in progress for this apiKey, return that.
```
rasendubi, line 363: The code below does not actually check that initialization is unique per api key

src/index.ts
```diff
@@ -252,26 +325,58 @@ export class EppoJSClient extends EppoClient {
...
+    if (!EppoJSClient.initializationPromise) {
+      EppoJSClient.initializationPromise = explicitInit(config, client).then((client) => {
+        // Resolve the ready promise if it exists
+        if (client.readyPromiseResolver) {
+          client.readyPromiseResolver();
+          client.readyPromiseResolver = null;
+        }
+        return client;
+      });
```
rasendubi, line 372: minor: now that `explicitInit` accepts a client as a parameter, would it make sense to make it a class member? it could call `readyPromiseResolver`, so we wouldn't need this hacky `initializeClient`.

I also think it was cleaner when `initializationPromise` stayed closer to global `init()`—`initializeClient()` was hard to follow because it takes a client instance but also accesses static fields (having that in `init()` was cleaner because `init()` knows that client instance stands for the singleton.)

src/index.ts
rasendubi, line 467: minor: maybe I'm asking too much from this PR, but have you considered pushing more of this initialization stuff into common SDK? I imagine that initialization sequence is more or less the same between client/node/native, only differing in default parameters.

Singleton handling is also a good candidate for sharing.

src/index.ts
```diff
@@ -335,41 +442,31 @@ export function offlineInit(config: IClientConfigSync): EppoClient {
...
- * Tracks pending initialization. After an initialization completes, the value is removed from the map.
- */
-let initializationPromise: Promise<EppoClient> | null = null;
-
 /**
  * Initializes the Eppo client with configuration parameters.
  * This method should be called once on application startup.
  * If an initialization is in process, calling `init` will return the in-progress
  * `Promise<EppoClient>`. Once the initialization completes, calling `init` again will kick off the
  * initialization routine (if `forceReinitialization` is `true`).
```
rasendubi, line 454: unrelated to the PR but I just noticed that the field is called `forceReinitialize`
```suggestion
 * initialization routine (if `forceReinitialize` is `true`).
```

src/i-client-config.ts
```diff
@@ -134,19 +166,11 @@ export interface IClientConfig extends IBaseRequestConfig {
    */
   updateOnFetch?: ServingStoreUpdateStrategy;
 
-  /**
-   * A custom class to use for storing flag configurations.
-   * This is useful for cases where you want to use a different storage mechanism
-   * than the default storage provided by the SDK.
-   */
-  persistentStore?: IAsyncStore<Flag>;
+  // TODO: Add initial config (stringified IConfigurationWire) here.
```
rasendubi, line 169: minor: it's better to add a proper `Configuration` class instead of passing untyped/unvalidated strings

src/i-client-config.ts
```diff
@@ -165,4 +189,77 @@ export interface IClientConfig extends IBaseRequestConfig {
...
+};
+
+/**
+ * Config shape for client v2.
+ */
+export type IClientOptions = IApiOptions &
+  ILoggers &
+  IEventOptions &
+  IStorageOptions &
+  IPollingOptions;
```
rasendubi, line 251: Are these sub-types ever used separately? Having the type split like this increases the chance of getting name clashes on the fields (which typescript won't really warn us about)

src/i-client-config.ts
```diff
@@ -134,19 +166,11 @@ export interface IClientConfig extends IBaseRequestConfig {
...
-  persistentStore?: IAsyncStore<Flag>;
+  // TODO: Add initial config (stringified IConfigurationWire) here.
+};
 
-  /**
-   * Force reinitialize the SDK if it is already initialized.
-   */
-  forceReinitialize?: boolean;
 
-  /** Configuration settings for the event dispatcher */
```
rasendubi, line 149: This doc string is now missing

sameerank (commented):
Don't forget to address the "defined but never used" warnings! Oleksii raised good points so please address those. Otherwise lgtm

src/index.ts
felipecsl: what does "ready" mean in this context? adding comments is likely to help, also we should always do it for public APIs anyways

src/index.ts
felipecsl: what does it mean and how is that different than this? consider renaming either if so

src/index.ts
```diff
@@ -252,26 +325,58 @@ export class EppoJSClient extends EppoClient {
...
- * Builds a storage key suffix from an API key.
- * @param apiKey - The API key to build the suffix from
- * @returns A string suffix for storage keys
- * @public
- */
-export function buildStorageKeySuffix(apiKey: string): string {
-  // Note that we use the first 8 characters of the API key to create per-API key persistent storages and caches
-  return apiKey.replace(/\W/g, '').substring(0, 8);
+  /**
+   * Tracks pending initialization. After an initialization completes, the value is removed from the map.
```
felipecsl, line 340: what map?

felipecsl (commented):
I like this directionally but doesn't seem like we're quite there yet. I might need to pull this branch locally to review more in depth once some of the current round of feedback is addressed

src/client-options-converter.ts
```diff
@@ -0,0 +1,44 @@
...
+    baseUrl: options.baseUrl,
+    requestTimeoutMs: options.requestTimeoutMs,
+    numInitialRequestRetries: options.numInitialRequestRetries,
+    numPollRequestRetries: options.numPollRequestRetries,
+    pollAfterSuccessfulInitialization: options.pollAfterSuccessfulInitialization,
+    pollAfterFailedInitialization: options.pollAfterFailedInitialization,
+    pollingIntervalMs: options.pollingIntervalMs,
+    throwOnFailedInitialization: options.throwOnFailedInitialization,
+    skipInitialPoll: options.skipInitialRequest,
+  };
```
aarsilv, line 41: This seems like it could e a pain to keep in sync. If these parameters are just a supserset, could we use the spread operator instead?
```ts
const parameters = {...options, eventDispatcher, isObfuscated: true}
```

Side note: I've heard a couple of places they'd like to be able to tell client _not_ to obfuscate, such as in development environments. So I wonder if we take this opportunity to make `isObfuscated` configurable.

src/i-client-config.ts
```diff
@@ -165,4 +189,77 @@ export interface IClientConfig extends IBaseRequestConfig {
...
+
+  /**
+   * A custom class to use for storing flag configurations.
+   * This is useful for cases where you want to use a different storage mechanism
+   * than the default storage provided by the SDK.
+   */
+  persistentStore?: IAsyncStore<Flag>;
+};
+
+export type IPollingOptions = {
```
aarsilv, line 208: oooh I like this additional organization

src/i-client-config.ts
```diff
@@ -165,4 +189,77 @@ export interface IClientConfig extends IBaseRequestConfig {
...
+  ILoggers &
+  IEventOptions &
+  IStorageOptions &
+  IPollingOptions;
+
+/**
+ * Configuration for regular client initialization
+ * @public
+ */
+export type IClientConfig = Omit<IClientOptions, 'sdkKey' | 'offline'> &
```
aarsilv, line 257: I think the omit/pick makes it clear what is happening

src/index.spec.ts
```diff
@@ -382,6 +394,167 @@ describe('EppoJSClient E2E test', () => {
...
+    });
+
+    afterEach(() => {
+      jest.restoreAllMocks();
+    });
+
+    it('should be independent of the singleton', async () => {
+      const apiOptions: IApiOptions = { sdkKey: '<MY SDK KEY>' };
+      const options: IClientOptions = { ...apiOptions, assignmentLogger: mockLogger };
+      const isolatedClient = new EppoJSClient(options);
```
aarsilv, line 436: do we want people calling constructors or should we have some method that does this? My spidey senses tell me the later, but that may be left over bias from the Java community's war on constructors a decade ago

src/index.spec.ts
```diff
@@ -382,6 +394,167 @@ describe('EppoJSClient E2E test', () => {
...
+    beforeAll(() => {
+      global.fetch = jest.fn((url: string) => {
+        callCount++;
+
+        const urlParams = new URLSearchParams(url.split('?')[1]);
+
+        // Get the value of the apiKey parameter and serve a specific variant.
+        const apiKey = urlParams.get('apiKey');
+
+        // differentiate between the SDK keys by changing the variant that `flagKey` assigns.
```
aarsilv, line 490: 👍

src/index.spec.ts
```diff
@@ -382,6 +394,167 @@ describe('EppoJSClient E2E test', () => {
...
+        const ufc = { flags: { [obfuscatedFlagKey]: flagConfig } };
+
+        return Promise.resolve({
+          ok: true,
+          status: 200,
+          json: () => Promise.resolve(ufc),
+        });
+      }) as jest.Mock;
+    });
+    afterAll(() => {
```
aarsilv, line 519: minor/linter but consider adding a space between this function and the one above

src/index.spec.ts
```diff
@@ -382,6 +394,167 @@ describe('EppoJSClient E2E test', () => {
...
+      jest.restoreAllMocks();
+    });
+
+    it('should be independent of the singleton', async () => {
+      const apiOptions: IApiOptions = { sdkKey: '<MY SDK KEY>' };
+      const options: IClientOptions = { ...apiOptions, assignmentLogger: mockLogger };
+      const isolatedClient = new EppoJSClient(options);
+
+      expect(isolatedClient).not.toEqual(getInstance());
+      await isolatedClient.waitForReady();
```
aarsilv, line 439: I may be late here but I wonder if we should call _this_ "initialize()" to make it more similar to the standalone "init" 🤔

src/index.spec.ts
```diff
@@ -382,6 +394,167 @@ describe('EppoJSClient E2E test', () => {
...
+          status: 200,
+          json: () => Promise.resolve(ufc),
+        });
+      }) as jest.Mock;
+    });
+    afterAll(() => {
+      jest.restoreAllMocks();
+    });
+
+    it('should operate in parallel', async () => {
```
aarsilv, line 523: 👍

src/i-client-config.ts
aarsilv: These types are so similar is it worth having separate ones for the constructor vs. init? What if we just allow sdkKey or apiKey for backward compatibility? I worry having two constructor input types could be confusing

src/index.ts
```diff
@@ -85,6 +107,7 @@ export {
 export { ChromeStorageEngine } from './chrome-storage-engine';
 
 // Instantiate the configuration store with memory-only implementation.
+// For use only with the singleton instance.
 const flagConfigurationStore = configurationStorageFactory({
```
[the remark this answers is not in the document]
aarsilv: More specifically, it's used as a stand-in singleton until initialization is complete

src/index.ts
```diff
@@ -98,22 +121,72 @@ const memoryOnlyPrecomputedBanditsStore = precomputedBanditStoreFactory();
...
+  private readyPromiseResolver: (() => void) | null = null;
+
   public static instance = new EppoJSClient({
     flagConfigurationStore,
     isObfuscated: true,
   });
-  public static initialized = false;
+  initialized = false;
+
+  constructor(optionsOrConfig: EppoClientParameters | IClientOptions) {
```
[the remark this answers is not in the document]
aarsilv: +1

src/index.ts
```diff
@@ -98,22 +121,72 @@ const memoryOnlyPrecomputedBanditsStore = precomputedBanditStoreFactory();
...
+    if (!v2Constructor) {
+      // Original constructor path; passthrough to super is above.
+      // `waitForReady` is a new API we need to graft onto the old way of constructing first and initializing later.
+
+      // Create a promise that will be resolved when EppoJSClient.initializeClient() is called
+      this.readyPromise = new Promise((resolve) => {
+        this.readyPromiseResolver = resolve;
+      });
+    } else {
+      this.readyPromise = explicitInit(
```
aarsilv, line 170: what happens if an error is hit? I think we may need outer try-catch here to respect our throw on failed initialization flag. And if that is false (which we hope to make the new default) we'll probably want to return a stand-in client similar to what the singleton has.

src/index.ts
aarsilv: agreed

src/i-client-config.ts
```diff
@@ -165,4 +189,77 @@ export interface IClientConfig extends IBaseRequestConfig {
      */
     maxQueueSize?: number;
   };
+};
+
+export type IStorageOptions = {
```
aarsilv, line 194: I think storage uniqueness is based on SDK which _should_ be fine for this use case as I don't imagine multiple instances all using same SDK key, but I wonder if we should include that in the documentation somewhere

aarsilv (commented):
Thanks for getting after this! I left a bunch of comments and have some higher-level musings as well. You're much closer to this problem than I am so much of what I say may not make sense 😅 

1. Agree with @rasendubi that I'm thinking much of this could actually be pushed down to the common package.
2. I'm having trouble seeing the win of having two sets of initialization parameters
3. I don't think we want people actually newing up the client--one of the reasons we did the singleton pattern was to avoid people shooting themselves in the foot. Plus constructors that do "work" make testing hard and whatnot
4. We seem to be putting the responsibility of keeping track of the client on the user. What if we basically kept most things the same, but had an optional "namespace" parameter for initialization and we stored the map (as well as returned the instance on init).

Normal flow:
```ts
await init(config);
getInstance().getBooleanAssignment(...); // true
```
"parallel flow" cartoon example:
```ts
await init(config);
await init({...config, namespace: 'another'});
getInstance().getBooleanAssignment(...); // true
getInstance('another').getBooleanAssignment(...); // false
```
Then implementing customers don't need to pass an instance everywhere, just a string namespace, which could be some shared constant or something.

But if they want to pass an instance around, they of course always could with the return value:
```ts
const defaultSingleton = await init(config);
const anotherSingleton = await init({...config, namespace: 'another'});
defaultSingleton.getBooleanAssignment(...); // true
anotherSingleton.getBooleanAssignment(...); // false
```

The stand-in client could be returned for any `getInstance()` without a ready instance so things don't explode if something goes wrong.

src/i-client-config.ts
rasendubi: I would actually keep `apiKey` and save `sdkKey` rename for the next major bump

src/index.spec.ts
rasendubi: The main reason to prefer builder methods is that they make it easier to maintain backward compatibility. If constructor is private, we can fundamentally change how construction works without breaking API. We can also add builder methods for common setups and give them meaningful names

src/index.spec.ts
rasendubi: This method doesn't actually initialize anything — it just waits till the client fetches the configuration and is ready to serve assignments

rasendubi: This is a good sign that we need better name or documentation for the method ^

src/index.ts
```diff
@@ -98,22 +121,72 @@ const memoryOnlyPrecomputedBanditsStore = precomputedBanditStoreFactory();
...
+            optionsOrConfig, // The IClientOptions wrapper
+            optionsOrConfig.flagConfigurationStore ?? // create a new, memory only FCS if none was provided
+              configurationStorageFactory({
+                forceMemoryOnly: true,
+              }),
+            optionsOrConfig.eventIngestionConfig // Create an Event Dispatcher if Event Ingestion config was provided.
+              ? newEventDispatcher(optionsOrConfig.sdkKey, optionsOrConfig.eventIngestionConfig)
+              : undefined,
+          ) // "New" construction technique; isolated instance.
+        : (optionsOrConfig as EppoClientParameters), // Legacy instantiation of singleton.
```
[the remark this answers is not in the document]
typotter: obviated by other changes.

src/index.ts
```diff
@@ -98,22 +121,72 @@ const memoryOnlyPrecomputedBanditsStore = precomputedBanditStoreFactory();
...
+      this.readyPromise = new Promise((resolve) => {
+        this.readyPromiseResolver = resolve;
+      });
+    } else {
+      this.readyPromise = explicitInit(
+        convertClientOptionsToClientConfig(optionsOrConfig),
+        this,
+      ).then(() => {
+        return;
+      });
```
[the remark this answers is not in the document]
typotter: Thanks for the note. Refactored around this anyway

src/index.ts
typotter: good catch.

src/index.ts
typotter: Thanks for the perspective; switched to a builder method and really cleaned everything up

src/i-client-config.ts
typotter: We have the client returning a stringified `IConfigurationWire` so devs don't need to worry about serializing the config string.

src/i-client-config.ts
typotter: The idea is that a dev could build different options objects then combine them for initialization. If they don't need custom event and storage options, for example, they wouldn't have those configs.

src/client-options-converter.ts
typotter: removed

src/i-client-config.ts
typotter: fixed up. moved sdkKey to the next major

src/index.spec.ts
typotter: moved to a builder

src/index.ts
typotter: init handles catching exceptions on init and throwing/logging depending on the option

typotter (commented):
Thanks everyone for the great comments and perspectives.

Refactored back to a little closer to where we started; dropped all the dual-constructor/config stuff.

src/index.ts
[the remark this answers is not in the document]
aarsilv: May be moot based on this comment, but I think you could have a backwards-compatible type as well (e.g., one that is a union that allows for either apiKey or sdkKey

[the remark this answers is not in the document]
rasendubi: > The constructor is already public, so we needed to keep the EppoClientParameters in the constructor to avoid a major change.

My understanding is that constructor is only _technically_ public (similar to how we have **many** internal methods public) and nobody is calling it directly. At the very least, because parameters are unsafe to be set by users (e.g., `isObfuscated` or `sdkVersion` are not meant to be set by users). So I don't think we need to keep it backward-compatible and/or keep supporting old/unsafe options

src/i-client-config.ts
rasendubi: In my ideal world, we have a `Configuration` class with `toString()` and `fromString()` methods. This way, devs still don't need to worry about serialization but they have a stronger hint at what they are expected to pass in

src/i-client-config.ts
rasendubi: Could you give an example of how you see it used in users' code?

src/index.ts
```diff
@@ -335,269 +611,46 @@ export function offlineInit(config: IClientConfigSync): EppoClient {
...
  * @public
  */
 export async function init(config: IClientConfig): Promise<EppoClient> {
-  validation.validateNotBlank(config.apiKey, 'API key required');
-
-  // If there is already an init in progress for this apiKey, return that.
-  if (!initializationPromise) {
-    initializationPromise = explicitInit(config);
+  if (initializationPromise === null) {
+    initializationPromise = getInstance().init(config);
```
rasendubi, line 639: minor: again, unrelated to the PR per se but I find this `initializationPromise` disturbing and a potential source of errors.

If `init()` is called with different configuration and `forceReinitialize: true`, the second call is going to be silently discarded. Which may lead to all sorts of troubles as client is now running with unexpected configuration.

With `forceReinitialize: false`, this handling will also silence a warning, so users have less opportunity to catch their bug.

At the very least, the case of concurrent initialization is worth a warning. But overall, I think it might be a good idea to handle `forceReinitialize` here in this function (as per my other comment)

src/i-client-config.ts
```diff
@@ -104,10 +105,41 @@ export interface IPrecomputedClientConfig extends IBaseRequestConfig {
...
+
+  /**
+   * Override the endpoint the SDK uses to load configuration.
+   */
+  baseUrl?: string;
+
+  /**
+   * Force reinitialize the SDK if it is already initialized.
+   */
+  forceReinitialize?: boolean;
```
rasendubi, line 124: As per my other comment, I propose to forbid re-initialization of standalone clients (because it's buggy and it's now possible to create new clients). While we can postpone the implementation of that, I believe that `buildAndInit` should not accept `forceReinitialize` already (as removing it later will be a major change)

rasendubi (commented):
Exposing `constructor` and `init()` method publicly is my only major concern.

The better handling of re-initialization is nice to have and we can implement it in future PRs (though it's a good idea to remove `forceInitialize` option now to avoid a breaking change in the future)

src/index.ts
```diff
@@ -102,18 +115,290 @@ export class EppoJSClient extends EppoClient {
...
+            'Eppo SDK is already initialized, skipping reinitialization since forceReinitialize is false.',
+          );
+          return this;
+        }
+      }
+
+      // If the instance was polling, stop.
+      this.stopPolling();
+      // Set up assignment logger and cache
+      this.setAssignmentLogger(config.assignmentLogger);
```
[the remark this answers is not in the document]
typotter: The force reinitialize is heavily used internally in tests to get around the singleton pattern, so it's super easy to get onboard with your proposal here -- deprecate and remove from the new builder

> The global init function should handle forceReinitialize by creating a new singleton and stopping the old one. This is to keep backward compatibility but also because it's the only case where this makes sense.

Wouldn't creating a new instance for the singleton make any existing reference to the singleton invalid?
The initialization routine swaps out the config stores and requestor; doesn't this prevent the potential overwrite? (If both configstores are using a persistent store keyed by SDK key, they'd overwrite into the persistent store, but then that's only read on initialization if it's not expired, so there is a corner case).

src/i-client-config.ts
typotter: Agreed.
Pushed the forceReinit check to the soon-to-be-deprecated global static init method. Also moved this option to `ICompatibilityOptions` (maybe better named `LegacyOptions`?)

src/i-client-config.ts
typotter: I have expounded upon this in `apiKey` docs

src/index.ts
```diff
@@ -102,18 +115,290 @@ export class EppoJSClient extends EppoClient {
...
   public static instance = new EppoJSClient({
-    flagConfigurationStore,
+    flagConfigurationStore: configurationStorageFactory({
+      forceMemoryOnly: true,
+    }),
     isObfuscated: true,
   });
-  public static initialized = false;
+
+  constructor(options: EppoClientParameters) {
```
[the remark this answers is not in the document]
typotter: Made the constructor private. 🥂 to progress

src/index.ts
```diff
@@ -102,18 +115,290 @@ export class EppoJSClient extends EppoClient {
...
+    return client;
+  }
+
+  /**
+   * Resolved when the client is initialized
+   * @private
+   */
+  private readonly initializedPromise: Promise<void>;
+
+  initialized = false;
```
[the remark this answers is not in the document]
typotter: done

src/index.ts
```diff
@@ -102,18 +115,290 @@ export class EppoJSClient extends EppoClient {
...
+
+  async init(config: IClientConfig): Promise<EppoJSClient> {
```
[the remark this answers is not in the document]
typotter: we do, but we can't since we need to call into it from the legacy global `init` method. I have marked with @internal and have tried to keep the difference between calling the global `init` and this class method minimal, yet with a reasonable separation of concerns. When we make a major bump, we can properly privatize this method and delete the global `init`.

src/index.ts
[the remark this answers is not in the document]
typotter: came as close to your proposal as possible; stayed short of swapping out the singleton instance, but I reason that the `fetchFlagConfigurations` does a proper job of stopping the existing polling and swapping out the config-requesting mechanisms to grab the new request parameters.
We do still have the issue of an in-flight request being able to overwrite after it's been `stopped`. 
I opened FF-3996 to track this as it's ultimately a bug in commons

src/index.ts
typotter: This "buffered" initialization was implemented because of massive amounts of traffic from one client. It is one of several measures taken to prevent the flood of requests from improper initialization of the eppo client (ex: calling init everywhere instead of `getInstance`).

I've added a warning here whenever `init` is called but the singleton instance is already initialized (or in progress).

src/index.ts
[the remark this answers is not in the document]
typotter: made the constructor private on the assumption that even though it was public, it was not used.

src/index.ts
typotter: The initialization is extra customized for js-client and much more straightforward for the other SDKs. I do want to unify the init as best as possible across the 4, but that's too much scope creep for this change.
Tracked in FF-3997

src/i-client-config.ts
typotter: There's a lot of refactoring to do around the configuration stores to properly plumb in Configuration and I can't wait.

src/client-options-converter.ts
typotter: https://github.com/Eppo-exp/js-sdk-common/pull/218

src/index.ts
typotter: `waitForConfiguration` to be in line with multiplatform's analog `PollerThread::waitForConfiguration`

src/i-client-config.ts
typotter: Working on some docs here:
https://github.com/Eppo-exp/eppo-docs/compare/tp/js/new-client?expand=1

src/index.ts
```diff
@@ -101,19 +102,72 @@ export class EppoJSClient extends EppoClient {
   // Ensure that the client is instantiated during class loading.
   // Use an empty memory-only configuration store until the `init` method is called,
   // to avoid serving stale data to the user.
+  /**
```
typotter, line 105: I am not sure why, but github does not seem to be showing the diff against typo/no-singleton / #170

typotter:
Several false starts and iterations later, we're ready for another review. Note: this PR is now stacked on #170 where changes are made just to how the singleton is referenced, leaving the focus of this PR on `waitForConfiguration` and `buildAndInit`.

src/i-client-config.ts
aarsilv: Perfect thanks! And now that I think about it if the have the same API it's actually probably not too bad if they are messing with eachother's persistent storage anyways as they should be using same configuration

src/index.ts
```diff
@@ -114,8 +116,52 @@ export class EppoJSClient extends EppoClient {
...
+    const client = new EppoJSClient({ flagConfigurationStore });
+
+    // init will resolve the promise that client.waitForConfiguration returns.
+    client.init(config);
+    return client;
+  }
+
+  /**
+   * Resolved when the client is initialized
+   * @private
```
aarsilv, line 134: super minor, but does this annotation really add anything given the `private` keyword two lines later?

src/index.ts
```diff
@@ -620,6 +657,7 @@ export async function init(config: IClientConfig): Promise<EppoJSClient> {
 /**
  * Used to access a singleton SDK client instance.
  * Use the method after calling init() to initialize the client.
+ * @deprecated
```
aarsilv, line 660: consider adding what the user should do instead after `@deprecated` (e.g., why is this deprecated)

docs/js-client-sdk.eppojsclient.initialized.md
```diff
@@ -4,6 +4,10 @@
 
 ## EppoJSClient.initialized property
 
+> Warning: This API is now obsolete.
```
felipecsl, line 7: did you mean `deprecated` instead of `obsolete`?

docs/js-client-sdk.getinstance.md
```diff
@@ -4,6 +4,10 @@
 
 ## getInstance() function
 
+> Warning: This API is now obsolete.
```
felipecsl, line 7: same question

src/i-client-config.ts
```diff
@@ -133,19 +164,9 @@ export interface IClientConfig extends IBaseRequestConfig {
...
-   * than the default storage provided by the SDK.
-   */
-  persistentStore?: IAsyncStore<Flag>;
-
-  /**
-   * Force reinitialize the SDK if it is already initialized.
-   */
-  forceReinitialize?: boolean;
-
+export type IEventOptions = {
```
felipecsl, line 169: this needs docs

src/i-client-config.ts
felipecsl: documentation for this type missing

src/i-client-config.ts
```diff
@@ -165,4 +186,75 @@ export interface IClientConfig extends IBaseRequestConfig {
...
+
+/**
+ * Configuration for regular client initialization
+ * @public
+ */
+export type IClientConfig = IApiOptions &
+  ILoggers &
+  IEventOptions &
+  IStorageOptions &
+  IPollingOptions;
```
felipecsl, line 260: i love this! thanks for organizing

docs/js-client-sdk.getinstance.md
felipecsl: also whenever deprecating something, we should always propose an alternative so users are not left guessing.

src/index.ts
```diff
@@ -109,10 +111,57 @@ export class EppoJSClient extends EppoClient {
     isObfuscated: true,
   });
 
+  /**
+   * @deprecated
```
felipecsl, line 115: alternative?

src/index.ts
```diff
@@ -109,10 +111,57 @@ export class EppoJSClient extends EppoClient {
...
+    // init will resolve the promise that client.waitForConfiguration returns.
+    client.init(config);
+    return client;
+  }
+
+  /**
+   * Resolved when the client is initialized
+   * @private
+   */
+  private readonly initializedPromise: Promise<void>;
```
felipecsl, line 138: nit: i'd rename to `initializationPromise` or simply `initPromise`

felipecsl (commented):
Main feedback is around documentation and providing alternative APIs for what's been deprecated. I'm ready to approve after that, great work!

docs/js-client-sdk.eppojsclient.initialized.md
typotter: I'm not sure where`yarn docs` is getting this text from

typotter: I am slo not sure why I can't get `yarn docs` to pick up the latest code and not document this as deprecated.

src/i-client-config.ts
typotter: added

src/i-client-config.ts
typotter: fixed

src/index.ts
typotter: undeprecated. I want to make a public `isInitialized`, but more work to the underlying config handling is needed first

src/index.ts
typotter: good point. It's not an initialized promise, it's a promise signalling when it is initialized.

src/index.ts
typotter: nope. eager IDE.

docs/js-client-sdk.getinstance.md
typotter: This file is auto-generated, so the text is out of my hands. I have, however, updated all of the deprecated tags with an alternative. thank you

typotter:
Thank you, @rasendubi @felipecsl @aarsilv for you reviews and patience as we explore this change.

I have made a key change - the singleton `init` and `getInstance` are no longer deprecated. I still fell that's the right direction to take, but for now, let's ship this feature and continue to polish afterwards.

src/index.ts
```diff
@@ -106,18 +108,64 @@ export class EppoJSClient extends EppoClient {
...
+   */
+  private readonly initPromise: Promise<void>;
+
+  /**
+   * Resolves the `initPromise` when initialization is complete
+   *
+   * Initialization happens outside the constructor, so we can't assign `initPromise` to the result
+   * of initialization. Instead, we call the resolver when `init` is complete.
+   */
+  private initializedPromiseResolver: () => void = () => null;
```
felipecsl, line 151: nit: `initPromiseResolver`

felipecsl (approved):

src/index.ts
typotter: ✔️

src/index.ts
typotter: this work has started in js-common 😄

PR #170, opened 2025-02-11
chore: move initialization into class methods, limit singleton use

typotter:
---
labels: mergeable
---
_Eppo Internal_
🎟️ towards FF-3888
📜 [Multiple Eppo Clients](https://www.notion.so/eppo/Multiple-concurrent-and-isolated-EppoClient-instances-running-within-an-application-18449cc0114380288eb3eaac580c4e53)
[Configuration Side-loading](https://www.notion.so/eppo/Configuration-side-loading-design-12249cc0114380bc9650d8dc4e189e77)

## Motivation and Context

**This PR is carved from #166 in an attempt to make #166 focused on the the "new" API.** This PR adds no net new functionality, however, the new `EppoJSClient.init` method, marked **@internal** does allow for unbuffered initialization calls just as the global `init` method previously allowed. This internal method can be made private once we remove the global `init` method in the next major release.

The `init` and `getInstance()` methods work on a singleton instance of the `EppoJSClient` which makes for a generally smooth developer experience given that the Eppo SDK essentially handles dependency management for the dev. This does have its own drawbacks, but those are not particularly relevant here.

There are use cases, however, when a non-singleton instance of the `EppoJSClient` is required. One such use case is embedding the Eppo SDK into a library which itself can be included in other applications. If that 3p library shipped with Eppo is integrated into another application that has the Eppo SDK running, there would be undefined behaviour as the SDK cannot handle this use case.

**TL;DR: We want to move away from singleton access and allow for multiple instance of the EppoJSClient while maintaining the current API**

## Description
- Move bulk of initialization to a class method
- funnel all references to the static `instance` through `getInstance()`
- deprecate `EppoJSClient.instance`
- make `initialized` and `ensureInitialized` class members; still set `EppoJSClient.initialized` as before to ensure backward compatibility with the current API

- doc comments

## How has this been tested?
- tests

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -109,9 +109,6 @@ export class EppoJSClient extends EppoClient {
     isObfuscated: true,
   });
 
-  /**
-   * @deprecated
-   */
   public static initialized = false;
 
   private initialized = false;
--- a/src/index.ts
+++ b/src/index.ts
@@ -271,6 +271,12 @@ export class EppoJSClient extends EppoClient {
     }
   }
 
+  /**
+   * Initialize the Eppo Client.
+   *
+   * @internal
+   * @param config
+   */
   async init(config: IClientConfig): Promise<EppoJSClient> {
     validation.validateNotBlank(config.apiKey, 'API key required');
     let initializationError: Error | undefined;
@@ -495,6 +501,12 @@ export class EppoJSClient extends EppoClient {
     return this;
   }
 
+  /**
+   * Initialize the client synchronously, for offline use.
+   *
+   * @internal
+   * @param config
+   */
   offlineInit(config: IClientConfigSync) {
     const isObfuscated = config.isObfuscated ?? false;
     const throwOnFailedInitialization = config.throwOnFailedInitialization ?? true;
--- a/src/index.ts
+++ b/src/index.ts
@@ -272,10 +272,7 @@ export class EppoJSClient extends EppoClient {
   }
 
   /**
-   * Initialize the Eppo Client.
-   *
    * @internal
-   * @param config
    */
   async init(config: IClientConfig): Promise<EppoJSClient> {
     validation.validateNotBlank(config.apiKey, 'API key required');
@@ -502,10 +499,7 @@ export class EppoJSClient extends EppoClient {
   }
 
   /**
-   * Initialize the client synchronously, for offline use.
-   *
    * @internal
-   * @param config
    */
   offlineInit(config: IClientConfigSync) {
     const isObfuscated = config.isObfuscated ?? false;
@@ -589,6 +583,7 @@ export function offlineInit(config: IClientConfigSync): EppoClient {
 
   // For backwards compatibility.
   EppoJSClient.initialized = true;
+
   return instance;
 }
 
@@ -619,6 +614,7 @@ export async function init(config: IClientConfig): Promise<EppoJSClient> {
 
   // For backwards compatibility.
   EppoJSClient.initialized = true;
+
   return client;
 }
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -573,6 +573,11 @@ export function offlineInit(config: IClientConfigSync): EppoClient {
 }
 
 /**
+ * Tracks pending initialization. After an initialization completes, this value is set to null
+ */
+let initializationPromise: Promise<EppoJSClient> | null = null;
+
+/**
  * Initializes the Eppo client with configuration parameters.
  * This method should be called once on application startup.
  * @param config - client configuration
@@ -591,10 +596,20 @@ export async function init(config: IClientConfig): Promise<EppoJSClient> {
       applicationLogger.warn(
         'Eppo SDK is already initialized, skipping reinitialization since forceReinitialize is false.',
       );
+      return instance;
     }
   }
 
-  const client = instance.init(config);
+  if (initializationPromise === null) {
+    initializationPromise = instance.init(config);
+  } else {
+    applicationLogger.warn(
+      'Initialization is already in progress. init should be called only once at application startup.',
+    );
+  }
+
+  const client = await initializationPromise;
+  initializationPromise = null;
 
   // For backwards compatibility.
   EppoJSClient.initialized = true;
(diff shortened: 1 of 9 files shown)


src/index.ts
```diff
@@ -287,308 +584,42 @@ export function buildStorageKeySuffix(apiKey: string): string {
  * @public
  */
 export function offlineInit(config: IClientConfigSync): EppoClient {
-  const isObfuscated = config.isObfuscated ?? false;
```
typotter, line 290: moved to class

src/index.ts
```diff
@@ -287,308 +584,42 @@ export function buildStorageKeySuffix(apiKey: string): string {
...
-    initializationPromise = explicitInit(config);
+    initializationPromise = instance.init(config);
   }
 
   const client = await initializationPromise;
   initializationPromise = null;
-  return client;
-}
-
-async function explicitInit(config: IClientConfig): Promise<EppoClient> {
```
typotter, line 369: moved to `EppoJSClient.init`

src/index.ts
```diff
@@ -252,15 +261,303 @@ export class EppoJSClient extends EppoClient {
...
     }
   }
+
+  /**
+   * Initialize the Eppo Client.
+   *
+   * @internal
+   * @param config
+   */
+  async init(config: IClientConfig): Promise<EppoJSClient> {
```
typotter, line 280: no changes here from `explicitInit`

src/index.ts
```diff
@@ -252,15 +261,303 @@ export class EppoJSClient extends EppoClient {
...
+    return this;
+  }
+
+  /**
+   * Initialize the client synchronously, for offline use.
+   *
+   * @internal
+   * @param config
+   */
+  offlineInit(config: IClientConfigSync) {
```
typotter, line 510: no changes from offline init

src/index.ts
```diff
@@ -287,308 +563,58 @@ export function buildStorageKeySuffix(apiKey: string): string {
...
-    }
-
-    if (initializationSource === ConfigSource.NONE) {
-      // Neither config loader produced useful config. Return error, if exists, in order from first to last: fetch, config store, generic.
-      initializationError = initFromFetchError
-        ? initFromFetchError
-        : initFromConfigStoreError
-          ? initFromConfigStoreError
-          : new Error('Eppo SDK: No configuration source produced a valid configuration');
+      applicationLogger.warn(
```
typotter, line 596: The diff makes it hard to see what's left in this method;

1) We keep the `forceReinitialize` check here to keep it out of the EppoJSClient.init method
2) Warn if an initialization is currently in progress

felipecsl:
Just a thought here, since this will require a major version bump, does it even make sense to deprecate the old APIs?

Since we're doing a major departure in terms of API design, it might not make sense to keep deprecated APIs around, as that will make the internal design a lot more complicated.

Since upgrading will require code changes anyways (assuming it will be a major bump), then we might as well just redesign it as a blank slate, not worrying about deprecations. WDYT?

src/index.ts
```diff
@@ -101,19 +101,28 @@ export class EppoJSClient extends EppoClient {
   // Ensure that the client is instantiated during class loading.
   // Use an empty memory-only configuration store until the `init` method is called,
   // to avoid serving stale data to the user.
+  /**
+   * @deprecated. use `getInstance()` instead.
+   */
```
aarsilv, line 106: Guessing the plan in the future is to make this private? Or have getInstance() construct just-in time if one not set?

src/index.ts
```diff
@@ -101,19 +101,28 @@ export class EppoJSClient extends EppoClient {
...
     flagConfigurationStore,
     isObfuscated: true,
   });
+
+  /**
+   * @deprecated
+   */
   public static initialized = false;
 
+  private initialized = false;
```
aarsilv, line 117: nice

src/index.ts
aarsilv: 🔥

aarsilv (approved):
Thanks for carving out this PR, makes things very clear. I like what I see! 🚀 

Consider throwing a `package.json` minor version bump in here just so you don't forget 😛

aarsilv (commented):
Thanks for iterating on all the comments!

The one thing I'm still confused about (probably because it's late 😛) is the deprecation of `getInstance()`. Are we expecting people to handle managing the instance they get from `buildAndInit()`? Does this mean that they would need to write their own `getInstance()` methods and wrapper assignment methods with null checks that return default value if not initialized? If so, I think this may be unnecessary friction and we should think about how to make the 99% use case--a singleton in a code base where they are not concerned about conflicts--as friction-free as possible.

typotter (commented):
> Just a thought here, since this will require a major version bump, does it even make sense to deprecate the old APIs?
The intent for this PR was no more than a minor bump. Is there a breaking change to the API I overlooked? @felipecsl 

> 
> Since we're doing a major departure in terms of API design, it might not make sense to keep deprecated APIs around, as that will make the internal design a lot more complicated.

This I agree on, however, that's better left for #166. This PR doesn't change any if the current API, nor deprecates it (without an immediately available equivalent, ex: `getInstance` instead of `EppoJSClient.instance`)

> Since upgrading will require code changes anyways (assuming it will be a major bump), then we might as well just redesign it as a blank slate, not worrying about deprecations. WDYT?

Again, the intent of this PR is no breaking changes.
The current focus is to deliver on the concrete ask of multiple instances and leveraging the opportunity to to incrementally work towards the ideal interface.
Deprecating with a replacement doesn't require a code change but if one is made to avoid the annotation, it's a very simple change. Accumulating some @deprecateds while we work out what the ideal is seems like a good tradeoff to allow us to continue to ship improvements and bug fixes to the current major version.

typotter:
> The one thing I'm still confused about (probably because it's late 😛) is the deprecation of `getInstance()`. Are we expecting people to handle managing the instance they get from `buildAndInit()`? Does this mean that they would need to write their own `getInstance()` methods and wrapper assignment methods with null checks that return default value if not initialized? If so, I think this may be unnecessary friction and we should think about how to make the 99% use case--a singleton in a code base where they are not concerned about conflicts--as friction-free as possible.

@aarsilv no deprecation of `getInstance` in this PR. Will likely drop that in #166 so we can ship the customer-requested feature and continue discussion around the evolution of the SDK in a non-blocking way.

PR #179, opened 2025-03-05
chore: update to latest common.

typotter:
## Motivation and Context
A recent change to the common package moved the source-of-truth of obfuscation to config store metadata (via format). The EppoJS client and its tests require updating to ensure the format metadata is being used instead of the deprecated internal isObfuscated field. 

## Description
- update to latest common
- minor bump to `3.12.0`
- set default format metadata where needed
- update tests to use an obfuscated config file (common package caches whether config is obfuscated)
- docs

## How has this been tested?
- no functional changes


greghuels (approved):
LGTM!

PR #193, opened 2025-03-19
chore:update docs

typotter:


docs/js-client-sdk.iclientconfig.eventingestionconfig.md
```diff
@@ -1,18 +1,18 @@
...
 
 **Signature:**
 
 ```typescript
-eventTracking?: {
+eventIngestionConfig?: {
         batchSize?: number;
         deliveryIntervalMs?: number;
-        enabled?: boolean;
+        disabled?: boolean;
```
greghuels, line 15: I think `enabled` is the most recent option

greghuels (changes requested):
it should be "enabled", not "disabled"

docs/js-client-sdk.iclientconfig.eventingestionconfig.md
typotter: you're right. And it should still be called `eventTracking`, not `eventIngestionConfig`.
I don't know where eventIngestionConfig is coming from on my system, but these doc changes are showing up in all of my PRs

typotter:
I blew away all the caches and `dist` I could find, plus a `yarn install --force` for good measure, and I was able to get the right basis for the docs.

PR #211, opened 2025-04-21
Ability to specify configuration dates with offline init

aarsilv:
---
labels: mergeable
---
_Eppo Internal_
[//]: #  (Link to the issue or doc corresponding to this chunk of work)
🎟️ Fixes: [FF-4344 - Ensure JS Client offline init sets config published and fetched at](https://linear.app/eppo/issue/FF-4344/ensure-js-client-offline-init-sets-config-published-and-fetched-at)
👯 **Related PR:** [`eppo #12880`](https://github.com/Eppo-exp/eppo/pull/12880)

## Motivation and Context
Evaluation details include the dates the configuration was published and fetched. However, when offline initialized, these need to be provided.

## Description
Add the publish and fetch dates for the configuration as optional parameters for the synchronous client configuration, and then set them to the underlying offline init memory store

not needed; bugfix

## How has this been tested?
* Added a test case to `index.spec.ts`
  image
* Installed this branch in local eppo environment; used the functionality to pass dates
  image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@eppo/js-client-sdk",
-  "version": "3.15.1",
+  "version": "3.16.0",
   "description": "Eppo SDK for client-side JavaScript applications",
   "main": "dist/index.js",
   "files": [
--- a/src/index.ts
+++ b/src/index.ts
@@ -66,6 +66,10 @@ export interface IClientConfigSync {
   enableOverrides?: boolean;
 
   overridesStorageKey?: string;
+
+  configPublishedAt?: string; // ISO Datetime String
+
+  configFetchedAt?: string; // ISO Datetime String
 }
 
 export { IClientConfig, IPrecomputedClientConfig };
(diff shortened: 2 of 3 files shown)


js-client-sdk.api.md
```diff
@@ -212,7 +212,9 @@ export interface IPrecomputedClientConfigSync {
     throwOnFailedInitialization?: boolean;
 }
 
-// @public (undocumented)
+// Warning: (ae-internal-missing-underscore) The name "NO_OP_EVENT_DISPATCHER" should be prefixed with an underscore because the declaration is marked as @internal
```
typotter, line 215: Lint warning via doc generation. fun. really we should be moving all the declarations out of index.ts and have only the exports.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@eppo/js-client-sdk",
-  "version": "3.15.0",
+  "version": "3.15.1",
```
typotter, line 3: Since the API surface is expanding, should this necessitate a minor bump?

src/index.ts
```diff
@@ -66,6 +66,10 @@ export interface IClientConfigSync {
   enableOverrides?: boolean;
 
   overridesStorageKey?: string;
+
+  configPublishedAt?: string; // ISO Datetime String
```
typotter, line 70: The DX of having to drill into the UFC for `offlineInit` is poor, but this is the only way currently and there's already precedent for passing around the flag entities directly instead of a wrapped config object. `ConfigWire` will solve this, so this is an ok workaround for now.

typotter (approved):

package.json
aarsilv: oh yeah good call

src/index.ts
aarsilv: I was just thinking how `ConfigWire` should make this all way better

PR #224, opened 2025-05-23
Provide rehydration hook for Isolatable Hybrid Configuration Store

aarsilv:
_Eppo Internal:_
🎟️ **Ticket:** [FF-4432 - JS Client SDK - Provide rehydration hook for use with IsolatedHybridConfigurationStore](https://linear.app/eppo/issue/FF-4432/js-client-sdk-provide-rehydration-hook-for-use-with)
🗨️ **Slack:** ["...there’s no way for us to then push into the MemoryStore that is used for assignments..."](https://eppo-group.slack.com/archives/C04G8MDTM26/p1747883932774819?thread_ts=1747880908.749629&cid=C04G8MDTM26)

## Motivation and Context
In certain situations it can be useful to manually trigger a rehydration of an `IsolatableHybridConfigurationStore`, either from the persistent store or from explicitly provided values.

For example, if you didn't want flags to update mid-user session, but have a way to detect when a user is idle and opportunistically update the flags then from a previously stored configuration.

## Description
To achieve this, we can modify the `IsolatableHybridConfigurationStore` to check for an optional `registerRehydrate()` method on its persistent store. If so, it will call that method passing a hook to a rehydration method. That hook can be called as-needed by the persistent store.

Has not yet, but as a follow up we can expand documentation in the [JavaScript Configuration caching docs](https://docs.geteppo.com/sdks/client-sdks/javascript/Initialization/#configuration-caching).

## How has this been tested?
Expanded `isolatable-hybrid.store.spec.ts`
image


docs/js-client-sdk.iclientconfigsync.configfetchedat.md
```diff
@@ -0,0 +1,11 @@
+<!-- Do not edit this file. It is automatically generated by API Documenter. -->
+
+[Home](./index.md) &gt; [@eppo/js-client-sdk](./js-client-sdk.md) &gt; [IClientConfigSync](./js-client-sdk.iclientconfigsync.md) &gt; [configFetchedAt](./js-client-sdk.iclientconfigsync.configfetchedat.md)
+
+## IClientConfigSync.configFetchedAt property
```
aarsilv, line 5: It seems updated docs from earlier work were not checked in

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@eppo/js-client-sdk",
-  "version": "3.16.1",
+  "version": "3.17.0",
```
aarsilv, line 3: New functionality so opted for minor version bump

src/isolatable-hybrid.store.ts
```diff
@@ -2,6 +2,15 @@ import { HybridConfigurationStore, IAsyncStore, ISyncStore } from '@eppo/js-clie
 
 export type ServingStoreUpdateStrategy = 'always' | 'expired' | 'empty';
 
+/**
+ * Extension of AsyncStore where its aware it's paired with a SyncStore.
+ * Useful when providing custom persistent stores.
+ */
+export interface IPairedAsyncStore<T> extends IAsyncStore<T> {
```
aarsilv, line 9: I welcome suggestions for better interface name 😅

src/isolatable-hybrid.store.ts
```diff
@@ -2,6 +2,15 @@ import { HybridConfigurationStore, IAsyncStore, ISyncStore } from '@eppo/js-clie
...
+  // If available, will be called when HybridConfigurationStore is constructed to provide a rehydration hook for the serving store
+  registerRehydrate?(rehydrate: (entries?: Record<string, T>) => Promise<void>): void;
```
aarsilv, line 11: Rather than a generic `registerHooks()` type method, I think we can add hooks one top-level method at a time. It won't make the flow control much more complex, and keep things cleaner I think.

src/isolatable-hybrid.store.spec.ts
```diff
@@ -130,4 +133,78 @@ describe('IsolatableHybridConfigurationStore', () => {
...
+        },
+      };
+
+      store = new IsolatableHybridConfigurationStore(syncStore, persistentStore, 'expired');
+    });
+
+    it('Provides hook to update serving store from persistent store', async () => {
+      await store.init();
+      // Expect sync store to be initialized with the async store
+      expect((await persistentStore.entries())?.['a']).toBe(initialEntries['a']);
```
aarsilv, line 181: at first I used the hard-coded values but I think it actually easier to follow when deferencing `initialEntries` and `newEntries` so you don't have to remember `1` is old and `3` is new.

aarsilv:
@pzuraq here is the WIP result of our pairing earlier today--please let us know if this will address your use case!

src/isolatable-hybrid.store.ts
typotter: agreed - we're not looking at needing an expandable hooks/plugin feature

src/isolatable-hybrid.store.ts
typotter: `RehydratingAsyncStore`

src/isolatable-hybrid.store.ts
```diff
@@ -39,8 +53,25 @@ export class IsolatableHybridConfigurationStore<T> extends HybridConfigurationSt
...
+      this.setServingStoreEntries(entries);
     }
     return updateServingStore;
   }
+
+  /**
+   * Hook to rehydrate serving store entries either from the persistent store or explicitly.
+   * Useful for manually triggering a serving store update when the upstate strategy is not set to always.
+   */
+  private async rehydrate(entries?: Record<string, T>): Promise<void> {
```
typotter, line 65: unfortunately, this is only partially rehydrating, as the config metadata is not being set. It's probably fine though for this use case

typotter (approved):

aarsilv:
It turns out this is not needed so going to close for now but consider the use case for future SDK development.

